PR #61, opened 2014-07-16
only activate agent if r10k is executable on $PATH

lmello:
This commit makes the node on mcollective discoverable
for the agent r10k only if the r10k binary is available
and executable on $PATH.

This way the agent will only run on machines that have
r10k installed.

This helper activate_when  only activate
this agent for discovery and execution if
the r10k binary is found on $PATH.
http://docs.puppetlabs.com/mcollective/simplerpc/agents.html#agent-activation

### DIFF
--- a/files/agent/r10k.rb
+++ b/files/agent/r10k.rb
@@ -1,6 +1,18 @@
 module MCollective
   module Agent
     class R10k<RPC::Agent
+       activate_when do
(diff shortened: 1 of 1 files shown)


acidprime:
@lmello I plan on adding this in 3.0 but need to rework the code as its potentially a breaking change for peeps

acidprime:
This has been pulled in

PR #78, opened 2014-09-12
Stash webhook helper

adamcrews:
This PR addresses https://github.com/acidprime/r10k/issues/62
I find it easier to switch to a different stash hook plugin than hack the Atlassian one.


igalic:
i'm a bit confused here, does _mco imply mcollective?

adamcrews:
@igalic yes, mco == mcollective.

PR #84, opened 2014-10-01
[Deprecation] Remove purgedirs support

trlinkin:
As of R10K 1.0.0 the purgedirs functionality and configuration options
have been removed. This module continues to configure the no longer
acknowledged configuration options.

This commit removes support for the purgedirs configuration option.

Oh, also... It should be mentioned that R10K 1.x was released over a year ago now. Reasonably, nobody should be running pre 1.x anymore.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -10,9 +10,6 @@
 # * [*sources*]
 #   Hash containing data sources to be used by r10k to create dynamic Puppet
 #   environments. Default: {}
-# * [*purgedirs*]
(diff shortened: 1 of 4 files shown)


trlinkin:
@acidprime Ok, hopefully this can make it into 3.0

I rebased against the latest commits and now it passes tests. I could not find any previous tests for purging and since this is a feature remove, I cannot imagine tests to be added. If you can think of tests that need to be added, I'll gladly whip them up.

PR #121, opened 2015-02-05
added metadata.json file for puppet versions 3.6 and older

vchepkov:
if r10k module is installed via Puppetfile/git PE 3.3 and older do not show version number. 
I converted existing Modulefile into metadata.json. Modulefile has been removed from puppetlabs modules for quite awhile now


acidprime:
Thanks!

PR #123, opened 2015-02-10
Added bind_address option to webhook.yaml

WhatsARanjit:
Some implementations default to listen only on localhost due to Rack/webrick handler.  Need an option to tell rack/webrick to listen on a certain address.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -33,6 +33,7 @@ class r10k::params
   # Webhook configuration information
   $webhook_user               = 'puppet'
   $webhook_pass               = 'puppet'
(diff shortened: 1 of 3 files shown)


manifests/params.pp
```diff
@@ -33,6 +33,7 @@
   # Webhook configuration information
   $webhook_user               = 'puppet'
   $webhook_pass               = 'puppet'
+  $webhook_bind_address       = $::ipaddress
```
elyscape: This should default to '0.0.0.0'. Otherwise the hook won't bind on all interfaces, as is the expected behavior.

PR #124, opened 2015-02-11
Use 0.0.0.0 as default webhook bind address

elyscape:
This resolves an issue caused by #123 wherein the webhook will accept connections only on the first network interface listed.  Setting 0.0.0.0 as the bind address makes the webhook accept connections on all interfaces.


acidprime:
good catch

PR #127, opened 2015-02-17
Install system_timer gem as directed by `r10k --help`

ghoneycutt:


ghoneycutt:
@acidprime You interested in this? If so, I'll write up the spec tests.

without system_timer you get the following when running r10k

`Faraday: you may want to install system_timer for reliable timeouts`

acidprime:
This is only on older versions of ruby right?

ghoneycutt:
I'm using Ruby 1.8.7 - The latest and greatest that EL6 supports :)

Tested with Ruby 1.9.3 and did not see the warning with system_timer missing. Tried to install it and the build blew up. Would wrapping it in a conditional based on `$::rubyversion` work?

acidprime:
A wrapper conditional  seems reasonable

acidprime:
Feel free to re-open if you want to still get this in

PR #143, opened 2015-03-03
Add commas in README example

cmurphy:
The prefixes example does not contain commas after the 'basedir'
parameters, so a new user copying the examples will run into a parse
error. This patch fixes the example.


acidprime:
Thanks

PR #147, opened 2015-03-16
fix for /var/run being mounted on tmpfs in ubuntu 14.04

spidersddd:
Ubuntu 14.04 is using a link to /run from /var/run.  The /run mount point is mounted on tmpfs which breaks the init scripts ability to start after a reboot.  This commit checks for the /var/run/webhood dir and creates it if it does not exist.


acidprime:
@spidersddd thanks for the contribution released in 2.6.5 on github and forge

PR #149, opened 2015-04-01
Support of Puppet FOSS for Webhook

mvisonneau:
I implemented the support of FOSS for the webhook. I currently only tested on Ubuntu 14.04.2 LTS
Some other enhancements : 
- Set GIT_SSL_NO_VERIFY if necessary
- Definition of a http_proxy to access remote repositories
- Custom mcollective cert files location

Example usage : 

```puppet
class { '::r10k::mcollective':
  http_proxy     => 'http://proxy.example.lan:3128',
  git_ssl_verify => 1,
}

class { '::r10k::webhook::config':
  enable_ssl       => false,
  protected        => false,
  public_key_path  => '/etc/mcollective/server_public.pem',  # Mandatory for FOSS
  private_key_path => '/etc/mcollective/server_private.pem', # Mandatory for FOSS
  notify           => Service['webhook'],
}

class { '::r10k::webhook':
  user    => 'puppet',                                       # Mandatory for FOSS
  group   => 'puppet',                                       # Mandatory for FOSS
  require => Class['::r10k::webhook::config'],
}
```

PR #150, opened 2015-04-08
Support of Puppet FOSS for Webhook

acidprime:
This is a fix for #149


mvisonneau:
Quicker than me :) Still working fine for me on Ubuntu 14.

templates/webhook.bin.erb
```diff
@@ -1,4 +1,4 @@
-#!/opt/puppet/bin/ruby
+#!<%= if @is_pe == true or @is_pe == 'true' then '/opt/puppet/bin' else '/usr/bin' end %>/ruby
```
igalic: this will fail when strict variables is enabled, cuz it doesn't actually exist…

mvisonneau: Can we add the following in the webhook.pp to solve this problem ?

```puppet
if ! defined( $is_pe ) {
  $is_pe = false
}
```

igalic: if i have [this discussion](https://github.com/puppetlabs/puppetlabs-concat/pull/270) another time, i'll scream :P

PR #151, opened 2015-04-15
Full path to r10k in postrun command

eskp:
When I was triggering puppet runs with mco puppet agent it wasn't finding r10k for some reason. Error message from puppet was "Could not execute posix command: No such file or directory". Manual runs were fine. Providing the full path seems to do the trick. Probably a good practice anyway.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -22,7 +22,7 @@ class r10k::params
   $remote     = "ssh://${git_server}${repo_path}/modules.git"
 
   # prerun_command in puppet.conf
-  $pre_postrun_command = 'r10k deploy environment -p'
+  $pre_postrun_command = '/usr/local/bin/r10k deploy environment -p'
 
   # Gentoo specific values
   $gentoo_keywords = ''


acidprime:
This is only the full path for FOSS not Puppet Enterprise though

acidprime:
Oh wait I guess I was creating a symlink and 3.8 and higher has that link, let me think a little more on this, I add the shell PATH to the agent, I might just want to append /usr/local/bin too instead

acidprime:
So actually I was creating that symlink for a while and now Puppet Enterprise 3.8 could but I would think other situations might be problematic. Given puppet supports setting the path via this configuration option: [path](https://docs.puppetlabs.com/references/latest/configuration.html#path) that would be the more generic way to fix this no?

acidprime:
@eskp one other way to fix this is declare the class with an overridden value for command:

https://github.com/acidprime/r10k/blob/master/manifests/postrun_command.pp#L3

```puppet
class r10k::postrun_command {
  command => '/usr/local/bin/r10k deploy environment -p',
}
```

Closing as I think the generic option is the correct one, but would be open to feedback to reopen if any of the above does not make sense

If this was not clear in the README I think a docs ticket would be better too

Cheers
-Z

acidprime:
@eskp Also postrun_command and the mco agent have nothing to do with each other normally. The mco agent has this code in 2.7.0 and higher https://github.com/acidprime/r10k/blob/master/templates/agent/r10k.rb.erb#L59

acidprime:
On puppet enteprise you should see something like this with 2.7.0 and higher of this module.


```shell
 grep -E '^.*environment.*PATH'  /opt/puppet/libexec/mcollective/mcollective/agent/r10k.rb 
        environment = {"LC_ALL" => "C","PATH" => "#{ENV['PATH']}:/opt/puppet/bin", "http_proxy" => "", "https_proxy" => "", "GIT_SSL_NO_VERIFY" => "0" }
```

acidprime:
If you using FOSS then take a look at this code to determine your path and please paste in this ticket 

https://github.com/acidprime/r10k/blob/master/manifests/params.pp#L81-L99

acidprime:
@eskp also this module ships and mcollective agent so you should not need to trigger puppet runs to trigger r10k , see this documentation https://github.com/acidprime/r10k#mcollective-support.
The `postrun_command` support was not intended as a way to bootstrap r10k via mcollective, then mcollective agent in this module does that, the postrun_comand was mainly so you did not have setup a cron job on your masters and just use the agent runs to do the normal checks. Most people don't use this any more as they tie their git server to r10k via the so it triggers syncs real time not on a schedule via something like the [webhook](https://github.com/acidprime/r10k#webhook-support)

eskp:
Makes sense @acidprime. Thanks for the quick and detailed reply. At the moment I'm monitoring my puppet control repository with Jenkins which triggers r10k and puppet runs with mcollective on the master. You are right, maybe I should move to webhook setup and take Jenkins out of the equation, however I do have FOOS Puppet. I did briefly have a look at mcollective r10k agent that comes with this module but could figure out how to trigger the run remotely from another machine (Jenkins).

PR #156, opened 2015-04-21
Add default install options

### LINKED ISSUE #155: Gem install has the space problem regressed.
See #87

binford2k:
This seems to be the default that most people want anyway, and it avoids
the recurring empty space gem install problem.

Fixes #155


acidprime:
This only works for gem installs, the module has multiple providers, looks like I need to break the package resource out

acidprime:
Will revist this during a redesign of the module with 3.x , but have a fix in the interim that allows all users to continue to work https://github.com/acidprime/r10k/commit/22fd541745a550a1aa46f220da0860bb222afb50

PR #158, opened 2015-04-28
Update r10k gem version for v2.7.x module version

aharden:
r10k gem v1.5.0 was yanked and params class defaults to v1.5.1


igalic:
can you properly [rebase](http://git-scm.com/book/en/Git-Branching-Rebasing) this?
this might be easier to do this with the [github flow](https://guides.github.com/introduction/flow/index.html) workflow.

Issue #160, opened 2015-05-01
use of ::pe_version fails puppet run when strict_variables is enabled on foss puppet


smithtrevor:
The logic here uses the pe_version fact which doesn't exist on foss.  This throws 'Unknown variable' error.

https://github.com/acidprime/r10k/blob/master/manifests/install.pp#L67

ghoneycutt:
@acidprime What do you think about pe_version in stdlib to prevent this?

rnelson0:
@smithtrevor What version of Puppet are you using? With 3.7.5 FOSS this compiles properly.
```
[root@puppet ~]# puppet --version
3.7.5
[root@puppet ~]# puppet apply -e "unless versioncmp($::pe_version, '3.8.0') >= 0 { notify {'less than 3.8.0': } }"
Notice: Compiled catalog for puppet.nelson.va in environment production in 0.06 seconds
Notice: less than 3.8.0
Notice: /Stage[main]/Main/Notify[less than 3.8.0]/message: defined 'message' as 'less than 3.8.0'
Notice: Finished catalog run in 0.14 seconds
```

smithtrevor:
@rnelson0 Originally it was on 3.7.5 and I get the same result on 3.7.3.  it's actually only an issue when strict_variables is set to true.  

```
[vagrant@cent7 ~]$ puppet --version
3.7.5
[vagrant@cent7 ~]$ puppet apply -e "unless versioncmp($::pe_version, '3.8.0') >= 0 { notify {'less than 3.8.0':} }"  --strict_variables true
Error: Undefined variable "::pe_version"; Undefined variable "pe_version" at line 1 on node cent7.local
Error: Undefined variable "::pe_version"; Undefined variable "pe_version" at line 1 on node cent7.local
```

A possible resolution to this would be to use getvar() in the conditional to test for existence of the fact prior to versioncmp()

```
[vagrant@cent7 ~]$ puppet apply -e "unless getvar('::pe_version') and versioncmp($::pe_version, '3.8.0') >= 0 { notify {'less than 3.8.0':} }"  --strict_variables true
Notice: Compiled catalog for cent7.local in environment production in 0.02 seconds
Notice: less than 3.8.0
Notice: /Stage[main]/Main/Notify[less than 3.8.0]/message: defined 'message' as 'less than 3.8.0'
Notice: Finished catalog run in 0.01 seconds
```

[...]

PR #161, opened 2015-05-01
add logic to prevent issue with pe_version on foss puppet with strict_variables enabled

smithtrevor:
Fix to #160 which prevents $::pe_version from being evaluated unless $provider is set to 'pe_gem'.


igalic:
:+1:!

smithtrevor:
This PR will probably change slightly based on the discussion for issue #160

PR #163, opened 2015-05-05
Add support for OpenBSD, some noteworthy points to mention:

buzzdeee:
- Tested with OpenBSD -current, where there is a r10k package available
  - Ruby binaries on OpenBSD are prefixed with the Ruby version they are built
    for, therefore r10k is called r10k21, since that is the version Puppet
    and r10k relies on
    - that would also be the case if r10k would be installed with the 'gem'
      package provider
  - the service name of mcollective is called 'mcollectived' on OpenBSD

Because of the last two mentioned points, added an osfamily dependent
 r10k_binary variable, as well as making the mc_service_name variable
osfamily dependent too.

What I tested was managing the installation of package and configuration
file /etc/r10k.yaml as well as the prerun script added to /etc/puppet/puppet.conf
file. Pre-run script comes in handy, since my modules/hieradata is on
a ram disk, and on boot, it auto-fills itself before a first puppet agent
run on the master.

Not tested is the mcollective integration, since I first have to
clean up my messed up mcollective setup. That's something that
would come later.

Let me know if there is something to add/fix/rework etc.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -20,6 +20,7 @@ class r10k::install (
 
   if $package_name == '' {
     case $provider {
+      'openbsd': { $real_package_name = 'ruby21-r10k' }
       'portage': { $real_package_name = 'app-admin/r10k' }
       'yum':     { $real_package_name = 'rubygem-r10k' }
       default:   { $real_package_name = 'r10k' }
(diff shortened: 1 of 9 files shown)


acidprime:
Any chance you can provide some rspec tests for this new os ? If not then I will get this merged as soon as I have time to generate them.

buzzdeee:
rspec tests added, I hope I didn't went overboard ;)
Only added specs for things I tested, i.e. no test with regard to mcollective or the webhooks, but those may follow when I get around to check it out.

PR #167, opened 2015-05-11
Fix a few typos.

fnaard:
This commit fixes a few tiny typos in the README.md file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,16 +108,16 @@ git_deploy_key { 'add_deploy_key_to_puppet_control':
   provider     => 'gitlab',
 }
 ```
-A simple example of creating a ssh private key would using an exec to call `yes y | ssh-keygen -t dsa -C "r10k" -f /root/.ssh/id_dsa -q -N ''`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,16 +108,16 @@ git_deploy_key { 'add_deploy_key_to_puppet_control':
   provider     => 'gitlab',
 }
 ```
-A simple example of creating a ssh private key would using an exec to call `yes y | ssh-keygen -t dsa -C "r10k" -f /root/.ssh/id_dsa -q -N ''`.
+A simple example of creating an ssh private key would use an exec to call `yes y | ssh-keygen -t dsa -C "r10k" -f /root/.ssh/id_dsa -q -N ''`.
 The example above shows using `git_deploy_key` which would deploy that key to the remote git server via its api. This is often required in the programtic creation of compile masters.
 
 Given r10k will likely be downloading your modules, often on the first server
-its ran on you will have to puppet apply this module to bootstrap this
+it's run on you will have to puppet apply this module to bootstrap this
```
ghoneycutt: Needs a comma, too. puppet apply should be in backticks like the commands above to show that it is a command.

it's run on, you will have to `puppet apply` this module to bootstrap this

PR #170, opened 2015-05-13
Fix non-ssl, abilliy to customize templates, and run with less privileges (but call r10k/mco via sudo)

robbat2:


templates/webhook.bin.erb
```diff
@@ -152,10 +156,10 @@ class Server < Sinatra::Base
           end
         else
           if $config['use_mcollective']
-              command = "mco r10k deploy #{branch} >> #{$config['mco_logfile']} 2>&1 &"
+              command = "#{$command_prefix} mco r10k deploy #{branch} >> #{$config['mco_logfile']} 2>&1 &"
```
acidprime: I am not super clear on why this is needed if you are running the script as the user, is the user case that you want to run the script as say root and then run r10k as non-root?

robbat2: I'm running webhook as non-root and sudoing to root for r10k. I modified all of the command calls for doing that. In the case of MCO, they'd run the webhook as nobody and sudo to puppet/peadmin for mco.

robbat2: After the template pull is in, I've got another branch ready for the Gentoo init script.
https://github.com/robbat2/r10k/commit/4dcba8212ead2fe3c829867e1fd8ecc6ce662c1f

robbat2:
The trailing commas accidentally added in the new block broke the webhook. There is a fix in my other PR.

PR #172, opened 2015-05-14
Gentoo & Pidfile/Init fixes

robbat2:
This started as writing a Gentoo init script, but ended up being fixing pidfiles so the init scripts could work properly.

### DIFF
--- a/templates/webhook.init.erb
+++ b/templates/webhook.init.erb
@@ -41,7 +41,6 @@ start() {
 
 stop() {
     echo -n $"Stopping webhook: "
-    pid=`cat $pidfile 2> /dev/null`
     killproc -p $pidfile $webhook
     RETVAL=$?
     return $RETVAL
@@ -54,7 +53,11 @@ restart() {
 
 status() {
     if [ -f "${pidfile}" ]; then
+    <% if @osfamily == 'Debian' -%>
+        status_of_proc -p $pidfile $webhook
+    <% else -%>
         status -p $pidfile $webhook
+    <% end %>
         RETVAL=$?
     else
         RETVAL=3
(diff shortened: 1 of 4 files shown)


robbat2:
Hmm, this might have introduced a subtle bug somewhere, and there isn't a testsuite for the webhook:
[2015-05-14 08:16:24] ERROR TypeError: no implicit conversion of Array into Integer
	/usr/lib64/ruby/2.0.0/openssl/ssl.rb:278:in `initialize'

robbat2:
Those trailing commas from me merging were the root of that webhook breakage.

templates/webhook.init.erb
```diff
@@ -41,15 +41,9 @@ start() {
...
-    pid=`netstat -lnp | awk  '/.*:8088.*ruby/{sub("/.*","",$NF);print $NF}'`
-    #killproc -p $pidfile $webhook]
-    if kill -9 $pid ; then
-      success 'webhook'
-      return 0
-    else
-      failure 'webhook'
-      return 1
-    fi
+    killproc -p $pidfile $webhook
```
acidprime: So this causes upgrade issues as the running demon does not have a pid

acidprime:
Also seeing 
```
service webhook status
/sbin/service: line 66: 12178 Segmentation fault      (core dumped) env -i PATH="$PATH" TERM="$TERM" "${SERVICEDIR}/${SERVICE}" ${OPTIONS}
```

templates/webhook.init.erb
robbat2: Hmm, how can we forcibly restart it just this once, so it has a pid in future?

robbat2:
I don't get the segfault on either Ubuntu 14.04 or Gentoo.

acidprime:
This is under Centos, trying to track it down

PR #180, opened 2015-05-25
Conditionally install the webhook packages

tampakrap:
Add a $manage_packages switch on the webhook class that enables the
installation of the webhook needed packages.
Move the installation of the packages to a r10k::webhook::package class
The reason is that we can skip installing the gems, and instead install distro
packages.

### DIFF
--- a/manifests/webhook/package.pp
+++ b/manifests/webhook/package.pp
@@ -1,0 +1,60 @@
+# Private class, do not include it directly.
+# Installs the webhook packages
+class r10k::webhook::package {
+  if $::is_pe == true or $::is_pe == 'true' {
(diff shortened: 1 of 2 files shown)


robbat2:
+1 for merging on this

manifests/webhook/package.pp
```diff
@@ -0,0 +1,60 @@
+# Private class, do not include it directly.
+# Installs the webhook packages
+class r10k::webhook::package {
+  if $::is_pe == true or $::is_pe == 'true' {
```
vchepkov: I find construct ``` if str2bool($::is_pe) { ```  better expresses the intent

PR #181, opened 2015-05-25
Updates in the webhook config

tampakrap:
* Fix the path of the configfile in the webhook binary
* Add support for /etc/webhook.yaml to be a symlink


templates/webhook.bin.erb
```diff
@@ -18,7 +18,7 @@ require 'yaml'
 require 'cgi'
 
 
-WEBHOOK_CONFIG = '/etc/webhook.yaml'
+WEBHOOK_CONFIG = '<%= @configfile %>'
```
tampakrap: this doesn't work because the configfile variable is set in `r10k::webhook::config` while the template is getting installed in `r10k::webhook`, I'll revert this one

PR #183, opened 2015-05-25
Include the r10k::webhook::config class in r10k::webhook

tampakrap:


acidprime:
@tampakrap please provide a description on why this is needed, typically the two classes are declared side by side.

tampakrap:
It's still possible to do it side by side if people are not using hieradata:

```puppet
class { 'r10k::webhook::config':
  user => 'puppet',
  group => 'pupet',
  ...
  enable_ssl => true,
  before => Class['r10k::webhook'],
}

class { 'r10k::webhook':
  manage_packages => false,
}
```

For users of hieradata, they will just need to `include r10k::webhook` instead of including `r10k::webhook::config` as well.

PS I was thinking of taking it one step further and introducing a $webhook param in the r10k class

acidprime:
Reverting this as its parse order dependant

tampakrap:
I apologize for this, but I don't quite understand the problem. Could you please give an example manifest so I can try to reproduce?

acidprime:
```puppet
class example ($message = 'foo bar'){
  notify { $message:}
}

include example

class {'example':
  message => 'I am declared after line 5',
}
```

```shell
Error: Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Class[Example] is already declared; cannot redeclare at /tmp/foo.pp:7 at /tmp/foo.pp:7:1 on node zacks-mbp.corp.puppetlabs.net
```

vs.

```puppet
class example ($message = 'foo bar'){
  notify { $message:}
}

class {'example':
  message => 'I am declared before line 5',
}

include example
```

```shell
: /Stage[main]/Example/Notify[I am declared after line 5]/message: defined 'message' as 'I am declared after line 5'
Notice: Applied catalog in 0.01 seconds
```

Basically you can't mix include and class resources declarations without them being parse order dependant

PR #189, opened 2015-06-10
change config location for version > 1.5.0

TheMeier:
R10K is released and removes /etc/r10k.yaml from the default search
path: https://tickets.puppetlabs.com/browse/RK-53

Change the config location to /etc/puppetlabs/r10k/r10k.yaml if veersion
is >= 1.5.0 and make sure directories are present. Also remove
/etc/r10k.yaml

fixes: #190

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -79,6 +79,17 @@ class r10k::config (
     $source_keys = keys($r10k_sources)
   }
 
+  if $configfile == '/etc/puppetlabs/r10k/r10k.yaml' {
+    file {['/etc/puppetlabs', '/etc/puppetlabs/r10k']:
+      ensure => directory,
+      before => File['r10k.yaml'],
+    }
+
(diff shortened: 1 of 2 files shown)


acidprime:
Sorry I have to to discuss this a little further the the Puppet Enterprise team before merging

acidprime:
So this is no longer the case, it was only the case for r10k 2.0

acidprime:
2.0.1 added this back in, I am not super comfortable managing this based on r10k version until 3.0 drops

TheMeier:
perfectly right for dropping this now but make some plans that won't happen again. from the upstream changelogs its quite clear that 2.X will support the legacy location but it's still deprecated

acidprime:
@TheMeier yeah its a bit of an akward situation as this module is still used by some PE customers but this will likely be the last version that supports PE, as soon as pe_r10k module has full support I will drop PE support and then can make changes that don't take PE into account

PR #193, opened 2015-06-19
Added support for BitBucket merge hooks

ayohrling:
Added support for git_provider configuration setting for the webhook to allow additional payload formats for the module and payload webhook endpoints. Defaulted to the current GitHub standard.

### DIFF
--- a/templates/webhook.bin.erb
+++ b/templates/webhook.bin.erb
@@ -79,7 +79,13 @@ class Server < Sinatra::Base
 
     data = JSON.parse(decoded, :quirks_mode => true)
 
-    module_name = ( data['repository']['name'] ).sub(/^.*-/, '')
+    if $config['git_provider'] == 'github'
+      module_name = ( data['repository']['name'] ).sub(/^.*-/, '')
+    elsif $config['git_provider'] == 'bitbucket'
+      module_name = (data['pullrequest_merged']['destination']['repository']).sub(/^.*-/, '')
+    else
+      $logger.error("Unsupport git provider: #{$config['git_provider']}")
+    end
 
     deploy_module(module_name)
   end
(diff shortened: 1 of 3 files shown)


templates/webhook.bin.erb
```diff
@@ -108,7 +114,13 @@ class Server < Sinatra::Base
     data = JSON.parse(decoded, :quirks_mode => true)
 
     # github sends a 'ref', stash sends an array in 'refChanges'
-    branch = ( data['ref'] || data['refChanges'][0]['refId'] ).sub('refs/heads/', '')
+    if $config['git_provider'] == 'github'
+      branch = ( data['ref'] || data['refChanges'][0]['refId'] ).sub('refs/heads/', '')
```
acidprime: This code actually is not just github, its github, gitlab and stash

ayohrling: Yeah I gathered that when I looked at my Gitlab environment, I needed a variable name to standardize on, so I chose GitHub as in "GitHub compliant," figured githublabstash would be too messy :) I am certainly open to suggestions of change.

acidprime: So Basically I would want to make this a required option for all services or fully automatic like it is not, My assumtion is you could just move your code `data['pullrequest_merged']['destination']['branch']['name']` to the end of the `||` conditionals and this would work out of the box without the need for the param to exist.

ayohrling: That would work. I was concerned with the logic clause becoming unmanageable over time. Either way actually works for me. I can resubmit.

ayohrling:
I updated the branch to merge in with just the inline evaluation for the JSON.

ayohrling:
Don't merge this yet, it doesn't work in practical testing. So it seems like the following is occurring; when the first attempt to access the JSON parsed output in the data variable and it's not found (because it's Bitbucket and doesn't have the GitHub format) ruby is throwing an exception and it's terminating the code at that point. I think we're going to have to add exception handling and rescue it out to the next parsing attempt for each body format.

ayohrling:
Confirmed, this is part of an issue with hash and nil items. Because data['repository'] returns nil, then data['repository']['name'] returns a NoMethodError and exits the code; the reason it worked for github vs. stash in the payload portion is that data['ref'] returns nil so it just goes to the next OR clause, in the case of BitBucket, the stash data access throws the NoMethodError and exits the code.

I am postulating the best way to handle this.

ayohrling:
Turns out I was able to just add an inline rescue that set nil in the event that it throws the error. This should work.

dharmabruce:
It appears this update will only handle bitbucket pull request merge events. I updated our local webhook code to handle bitbucket push events by referring to this JSON key for the branch name:
data['push']['changes'][0]['new']['name']
Could or should this be included or used instead? I verified a bitbucket pull request merge webhook event and a push webhook event are both sent after a merge.

acidprime:
@ayohrling if you can address @dharmabruce concern I will look at merging this

dharmabruce:
Tested PR submitted: https://github.com/acidprime/r10k/pull/227

PR #194, opened 2015-06-19
188 - Webhook segfault on status

ayohrling:
Addresses issue 188

When not debian or SUSE, the call to the status function is invalid.

Fixed the ordering of the status call to
status $webhook $pidfile

Also, collision between the internal status function and the status
fucntion within /etc/init.d/functions. Because of this segfault was
occurring. Renamed internal function to whstatus.

### DIFF
--- a/templates/webhook.init.erb
+++ b/templates/webhook.init.erb
@@ -53,14 +53,14 @@ restart() {
     start
 }
 
-status() {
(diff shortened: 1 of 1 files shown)


acidprime:
THANKS , I have not had a chance to track this down

PR #198, opened 2015-06-22
Restore EnvironmentFile line as optional

### LINKED ISSUE #197: EnvironmentFile removed from webhook unit file
In commit 5331cd9ac8291a1ea46d1b10981db3221db69308 the "EnvironmentFile=" line was removed, this was required to set environment variables like $GEM_PATH. In the original unit file EnvironmentFile wasn't marked option which would causes webhook not to start if the file wasn't present. The following link contains information on implementing EnvironmentFile as option, which is the best of both worlds.

https://fedoraproject.org/wiki/Packaging%3aSystemd#EnvironmentFiles_and_support_for_.2Fetc.2Fsysconfig_files

brandonweeks:
Fixes #197 by restoring the EnvironmentFile line as option. See below for documentation on unit files.

https://fedoraproject.org/wiki/Packaging%3aSystemd#EnvironmentFiles_and_support_for_.2Fetc.2Fsysconfig_files

### DIFF
--- a/templates/webhook.service.erb
+++ b/templates/webhook.service.erb
@@ -4,6 +4,7 @@ After=syslog.target network.target
 
 [Service]
 Type=forking
+EnvironmentFile=-/etc/sysconfig/webhook
 User=<%= @user %>
 ExecStart=/usr/local/bin/webhook
 PIDFile=/var/run/webhook/webhook.pid


acidprime:
Thanks for your contribution!

PR #212, opened 2015-07-06
Add postrun to r10k.yaml

### LINKED ISSUE #211: Feature Request: add a r10k postrun hook for directory environment cache invalidation
Hi there,

according to the docs, it is good common practice to set `environment_timeout = unlimited` in directory environments and on updates do things to tell the puppet master that code has changed.
(see http://docs.puppetlabs.com/puppet/4.1/reference/environments_configuring.html#environmenttimeout and older versions like http://docs.puppetlabs.com/puppet/3.8/reference/environments_configuring.html#environmenttimeout)

I would appreciate a script that handles this called by the r10k postrun. For now I would only enable the API call to Puppetserver (http://docs.puppetlabs.com/puppetserver/latest/admin-api/v1/environment-cache.html), but this could be extended to support passenger/rack deployments, too.

Shall I open a PR addressing this issue? Same thing with time slots as for #210... ;-)

poikilotherm:
Fixes #211.

This simply adds the `postrun` r10k configuration option to `r10k.yaml`.

You can choose a script for yourself, maybe provided by a profile etc.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -10,6 +10,10 @@
 # * [*sources*]
 #   Hash containing data sources to be used by r10k to create dynamic Puppet
 #   environments. Default: {}
+# * [*postrun*]
+#   **Optional:** Array containing the parts of a system call. 
+#   Example: ['/usr/bin/curl', '-F', 'deploy=done', 'http://my-app.site/endpoint']
+#   Default: undef
 # * [*manage_configfile_symlink*]
 #   Boolean to determine if a symlink to the r10k config file is to be managed.
 #   Default: false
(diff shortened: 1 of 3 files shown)


acidprime:
Specs failed, will take a look at this once they are passing

poikilotherm:
Thanks for commenting. I'm currently on vacation, thus not currently working on this. This is a dumb Ruby 1.8 issue, as 1.8 is not serializing arrays in the way newer version does (and I hadn't it on radar).

acidprime:
yeah this is the classic hash ruby 1.8.7 problem. I ended up not using `to_yaml` in the main ERB for this same reason.

acidprime:
@poikilotherm I have pulled this in a way that will allow ruby 1.8.7 to work , closing.

PR #214, opened 2015-07-07
Changes in r10k::webhook class

### LINKED ISSUE #213: webhook::package inclusion in r10k::webhook class
Currently, r10k::webhook::package class is included by using relative class path, which is not working anymore in Puppet 4.0 (and upper, as only "strict mode" there is enabled).

pdaukintis:
Include r10k::webhook::package by using full class path.
This allows to run this code on Puppet 4.x

Fixes #213

### DIFF
--- a/manifests/webhook.pp
+++ b/manifests/webhook.pp
@@ -49,7 +49,7 @@ class r10k::webhook(
   }
 
   if $manage_packages {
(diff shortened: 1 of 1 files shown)


johnzimm:
Thanks!  Just ran into this.

PR #218, opened 2015-07-17
(METHOD-401) Fix Webhook Service on EL7

jolshevski:
Previously, the webhook process was not terminated by systemd when
stopping the service. This caused reloading of the service to raise a
socket collision error due to the persistence of the old process.

This commit makes the appropriate systemd configuration changes to
facilitate stopping/restarting of the service.

### DIFF
--- a/templates/webhook.rehat.service.erb
+++ b/templates/webhook.rehat.service.erb
@@ -3,11 +3,13 @@ Description=R10K Webhook Service
 After=syslog.target network.target
 
 [Service]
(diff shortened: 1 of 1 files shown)


acidprime:
:+1:

acidprime:
@jolshevski how much testing did you do with this

There is this comment here:

https://github.com/acidprime/r10k/issues/207#issuecomment-116652501


That notes the user value too

PR #219, opened 2015-07-20
Allow module to work with PE 3.8.x Razor (no master)

esalberg:
- Only manage the peadmin cert if use_mcollective is true
 - Allow r10k to be installed for PE 3.8+ if puppet_master is set to false
 - Update rspec and readme

### DIFF
--- a/manifests/install/pe_gem.pp
+++ b/manifests/install/pe_gem.pp
@@ -1,11 +1,13 @@
 # This class links the r10k binary for PE
-class r10k::install::pe_gem {
+class r10k::install::pe_gem (
+  $puppet_master,
+) {
 
   require git
 
   # Puppet Enterprise 3.8 ships code to manage this symlink on install
   # This conditional should not effect FOSS customers based on the fact
-  unless versioncmp($::pe_version, '3.8.0') >= 0 {
+  unless versioncmp($::pe_version, '3.8.0') >= 0 and $puppet_master {
     file { '/usr/bin/r10k':
       ensure  => link,
       target  => '/opt/puppet/bin/r10k',
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ class { 'r10k':
 }
 ```
 _Note: On Puppet Enterprise 3.8 and higher the package is not declared as 3.8
-ships with an embdedded r10k gem installed via the PE packages_
+ships with an embdedded r10k gem installed via the PE packages. To install r10k on a PE 3.8+ non-master, set puppet_master to false for the main r10k class (e.g. to use r10k with Razor)_
 
 ## Usage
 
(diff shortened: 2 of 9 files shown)


esalberg:
I have this working (from basic testing) in my Razor test environment.  Please let me know if I need to make any changes to rspec, etc. - I'm not actually spec testing at the moment, so I didn't get a chance to review that part directly.

acidprime:
Can you provide a little background on what this solves related to razor? i.e. without this patch whats the symptom , looks pretty good just need the context

acidprime:
Ok re-read this, it looks like using the webhook without mcollective so you can trigger r10k to sync stuff to razor, seems reasonable to me

manifests/install/pe_gem.pp
```diff
@@ -1,11 +1,13 @@
...
+class r10k::install::pe_gem (
+  $puppet_master,
+) {
 
   require git
 
   # Puppet Enterprise 3.8 ships code to manage this symlink on install
   # This conditional should not effect FOSS customers based on the fact
-  unless versioncmp($::pe_version, '3.8.0') >= 0 {
+  unless versioncmp($::pe_version, '3.8.0') >= 0 and $puppet_master {
```
acidprime: I do need clarification on this one though, This check is for only nodes that are 3.7 and below so they don't declare this symlink as currently the puppet_enterprise module in 3.8 and higher manages the creation of the symlink. Can you provide a little more detail of what happen without this.

esalberg:
Sure!

There were two issues:

1) Needed to install the r10k gem (pe_gem provider)
The PE 3.8+ masters come bundled with r10k, but the razor server is just an agent with special additions, so it still needs the r10k gem installed.  Setting puppet_master to false bypasses the conditionals related to not installing r10k on PE 3.8+.

2) Remove dependency on peadmin-cert.pem
There was one place in the code where even though use_mcollective was set to false (because there's no peadmin user / mco client on a razor server), there was still a reliance on peadmin-cert.pem.  I just used the existing use_mcollective parameter to get rid of that dependency.

I also changed the user and pass in /etc/webhook.yaml to "pe-razor" - I have no idea if that was necessary, but I was troubleshooting the webhook and once I got it going, I didn't take the time to check into it further.  I'm guessing that setting protected to "false" was more important.  I can do a quick test with changing those values if you'd like - let me know.

(The certpath and client_cfg entries in /etc/webhook still reference /var/lib/peadmin items, but they don't seem to be causing any trouble.)

manifests/install/pe_gem.pp
esalberg: I was assuming that since r10k wasn't installed by puppet_enterprise that the symlink wouldn't be there as well.  I can test that out to validate.

acidprime:
Ah yeah that makes sense, as the r10k gem is not on the agent thus the master param

acidprime:
This looks good to go but needs to be rebased off of the updates needed for Puppet Enterprise 2015.2

acidprime:
I pulled this in https://github.com/acidprime/r10k/commit/cb5eb2d2191e92e8a0cff293b438663f43acdbae

esalberg:
I was just about to rebase - thanks for taking care of that for me. :)

acidprime:
@esalberg sorry I didn't pull in your commits in a bit of hurry today, will def give you a shout out in the CHANGELOG though

esalberg:
No worries, I'm not in it for the fame. :)

Issue #220, opened 2015-07-21
r10k and Webhook install using system ruby on CentOS 6 using Puppet 4.2.0 from "Puppet Collection 1" repo.


johnzimm:
CentOS 6 system ruby has version 1.8.7.  The current version of the webhook uses functions from the ruby stdlib that only exist after 1.9.2.  The ruby version installed with puppet in /opt/puppetlabs/puppet/bin/ruby is version 2.1.6p336.

Can using the ruby installed alongside puppet be accounted for in this module like it is for Puppet Enterprise?  pe_gem makes reference that it is deprecated and we should use puppet_gem, but I'm not finding puppet_gem to see if I can modify the r10k module to leverage that.

PR #221, opened 2015-07-21
Use included ruby and gem for Puppet FOSS 4.2.0

johnzimm:
Add support for puppet_gem provider to account for included ruby/gem in Puppet FOSS 4.2.0 and higher.

This is not well tested, but is working in my environment with Puppet FOSS 4.2.0.  It was done in a bit of a rush and I tried to model what was being done for the other provider methods without breaking anything.

Please take a look and suggest improvements.  Relates to issue #220 that I submitted.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -47,6 +47,11 @@ class r10k::install (
           version                => $version;
         }
       }
+      elsif $provider == 'puppet_gem' {
+        # Puppet FOSS 4.2 and up ships a vendor provided ruby.
+        # Using puppet_gem uses that instead of the system ruby.
+        include r10k::install::puppet_gem
+      }
       elsif $provider == 'pe_gem' {
         include r10k::install::pe_gem
       }
@@ -72,6 +77,7 @@ class r10k::install (
           install_options => $provider_install_options
         }
       }
+
     }
     default: { fail("${module_name}: ${provider} is not supported. Valid values are: 'gem', 'pe_gem', 'puppet_gem', 'bundle', 'openbsd', 'portage', 'yum', 'zypper'") }
   }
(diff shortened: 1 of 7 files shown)


manifests/install.pp
```diff
@@ -64,6 +67,7 @@
       }
 
       # Puppet Enterprise 3.8 and ships an embedded r10k so thats all thats supported
+      # Puppet FOSS 4.2 and up also ships an embedded r10k so thats all that should be supported
```
acidprime: This comment seems wrong, its not that it ships an embedded r10k , it ships a vendored ruby install without the r10k gem (thus the need to install below) to be clear

templates/webhook.bin.erb
```diff
@@ -1,4 +1,4 @@
-#!<%= if @is_pe == true or @is_pe == 'true' then '/opt/puppet/bin' else '/usr/bin' end %>/ruby
+#!<%= if @is_pe == true or @is_pe == 'true' then '/opt/puppet/bin' elsif @puppetversion >= '4.2.0' then '/opt/puppetlabs/puppet/bin' else '/usr/bin' end %>/ruby
```
acidprime: Can you let me know what OS you tested this with?

acidprime:
@johnzimm couple of minor comments but looks pretty good, if you can include spec updates I will merge immediately otherwise it will be a few days before I can get the tests updated after pulling in your commits.

manifests/install.pp
johnzimm: Yeah.  I copied the comment line.  I meant to change r10k to ruby.  Will get that changed.

templates/webhook.bin.erb
johnzimm: I only tested on CentOS 6.

johnzimm:
@acidprime I should be able to get updates done tomorrow morning.  Thanks!

johnzimm:
Updates made.  I'll rebase after I test them in the morning.

johnzimm:
Specs have been added.  I have not added specs before so I just copied the ones I felt were appropriate.  The Travis CI builds are failing because of r10k paths.  There may be other adjustments I have to make.  I should have time to look closer this evening.

acidprime:
I pulled this in with your commits, but your specs looked backwards to the logic, might want to take a look at the differences in the commits after yours

Issue #210, opened 2015-07-03
Feature Request: custom fact location of r10k binary


poikilotherm:
Hi there,

I would like to see a custom fact that sends the master the absolute path of the r10k binary.

My use case is the proper definition of a sudo rule for the webhook service user, where I need an absolute path (forced by sudo config enforcements).

Shall I create an PR addressing this issue? I can spend time on other things if you guys don't like the idea... ;-)

acidprime:
Yeah I would love this as I actually would want to use in the mcollective agent too

PR #222, opened 2015-07-28
Add a custom fact 'r10k_path'

poikilotherm:
Solves #210.

This will send a fact with the absolute path to the installed r10k binary to the master.

There is just one regression without an easy solution: the installed binary needs to be in the path of the user puppet agent is run as, which may be very different than the path the webhook / mco user uses.

### DIFF
--- a/lib/facter/r10k_path.rb
+++ b/lib/facter/r10k_path.rb
@@ -1,0 +1,11 @@
+Facter.add(:r10k_path) do
+  confine :kernel => :linux
+  setcode do
+    path = Facter::Util::Resolution.exec("which r10k")
+    if path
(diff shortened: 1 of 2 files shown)


lib/facter/r10k_path.rb
```diff
@@ -0,0 +1,11 @@
+Facter.add(:r10k_path) do
+  confine :kernel => :linux
+  setcode do
+    path = Facter::Util::Resolution.exec("which r10k")
```
acidprime: There are a few ruby examples of doing which so we don't then rely on this bindary, any interest in doing that to make it more platform independent?

PR #223, opened 2015-07-28
Puppet 4 future proof

acidprime:
Pulling in the Changes needed for the latest version of Puppet Enterprise

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -8,6 +8,8 @@ class r10k::install (
   $manage_ruby_dependency,
 ) {
 
+  $pe_server_version = $r10k::params::pe_server_version
+
   # There are currently bugs in r10k 1.x which make using 0.x desireable in
   # certain circumstances. However, 0.x requires make and gcc. Conditionally
   # include those classes if necessary due to 0.x r10k version usage. When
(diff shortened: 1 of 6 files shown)


manifests/install.pp
```diff
@@ -65,7 +67,7 @@
 
       # Puppet Enterprise 3.8 and ships an embedded r10k so thats all thats supported
       # This conditional should not effect FOSS customers based on the fact 
-      unless ($::is_pe == 'true' or $::is_pe == true) and versioncmp($::pe_version, '3.8.0') >= 0 {
+      unless $is_pe_server and versioncmp($::puppetversion, '3.8.0') >= 0 {
```
ghoneycutt: suggest using `if !` as `unless` is confusing

manifests/params.pp
```diff
@@ -57,8 +57,38 @@
     $functions_path     = '/etc/rc.d/init.d/functions'
     $start_pidfile_args = '--pidfile $pidfile'
   }
+  
   if $::is_pe == true or $::is_pe == 'true' {
-    # Puppet Enterprise specific settings
+    # < PE 4
+    $is_pe_server      = true
+  }elsif is_function_available('pe_compiling_server_version') {
```
ghoneycutt: why the extra check here? is `is_pe` not to be trusted?

acidprime: @ghoneycutt its going away as there is no such thing as PE in Puppet 4 on the agent side as its an all in one installer now

ghoneycutt: Interesting.. so there will be no difference between the FOSS and PE packages and no need to tell the difference from that standpoint? If there is no reason for `is_pe` anymore, why do you bother setting it in this code? Apologies if my comments are getting sidelined from the PR and I appreciate your explanations of the pending fact changes.

acidprime:
Closing in liue https://github.com/acidprime/r10k/pull/224

PR #227, opened 2015-07-30
Added support for BitBucket push webhooks.

dharmabruce:
This update replaces the BitBucket merge pull request event webhook handler bringing it to feature parity with the github webhook handler. I've tested that this update works for both BitBucket and github repos and that pull request merges in BitBucket result in push webhook events (as you would expect).


acidprime:
Have you tested this in the negative though , seems like `has_key?` would be safer

dharmabruce:
I additionally tested:
. The deletion of a bitbucket branch.

PR #229, opened 2015-07-31
Making r10k webhook 2015.x compatible

WhatsARanjit:
Testing that the correct packages are in the catalog when installing r10k webhook.  Satisfying the following scenarios:
 
* PE >=2015.x with webhook
* PE Puppet <v4.2.0, >=v3.7.0 with webhook
* PE Puppet <v3.7.0 with webhook
* FOSS >=2015.x with webhook
* FOSS Puppet <v4.2.0, >=v3.7.0 with webhook
* FOSS Puppet <v3.7.0 with webhook

### DIFF
--- a/spec/classes/webhook/package_spec.rb
+++ b/spec/classes/webhook/package_spec.rb
@@ -17,19 +17,19 @@ describe 'r10k::webhook::package' , :type => 'class' do
     end
     it { should contain_package('sinatra').with(
         'ensure'   => 'installed',
-        'provider' => 'puppet_gem',
+        'provider' => 'puppet_gem'
       )
     }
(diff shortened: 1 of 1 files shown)


spec/classes/webhook/package_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+      }
+    end
+    let :params do
+      {
+        :is_pe_server => true,
+      }
+    end
+    it { should contain_package('sinatra').with(
+        'ensure'   => 'installed',
+        'provider' => 'pe_gem',
```
acidprime: This needs to be changed for ruby 1.8.7

PR #233, opened 2015-08-04
Let rb scripts resolve what ruby to use themselves

attachmentgenie:
when creating rpm packages for this module, the resulting rpm complains that our systems dont have the required /opt/puppet/bin/ruby.
Its nicer to let the file/system decide what ruby to use


acidprime:
@attachmentgenie out of curiosity are you rpm'ing this for an OS distro or just internally?

attachmentgenie:
@acidprime packaging internally as part of a puppetmaster bootstrap project

PR #235, opened 2015-08-05
Add the ability to respond only to certain events

binford2k:
This adds a new webhook param, which allows you to define an event or
list of events to respond to as a single string or an array.

``` Puppet
class {'r10k::webhook::config':
  enable_ssl        => true,
  protected         => true,
  repository_events => ['push','release'],
  notify            => Service['webhook'],
}
```

This required extending the `webhook.yaml.erb` to also expand and sort
array data types.

Yes, you could do the same by only configuring certain github events,
but this allows you to more easily do it in code on the master.

Limitations: currently supports only github events.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -46,7 +46,8 @@ class r10k::params
   $webhook_r10k_deploy_arguments = '-pv'
   $webhook_bin_template          = 'r10k/webhook.bin.erb'
   $webhook_yaml_template         = 'r10k/webhook.yaml.erb'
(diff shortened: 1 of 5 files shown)


binford2k:
Btw, I didn't muck with the spec tests, because they all fail on master for me.

acidprime:
Try `bundle exec rake spec` <- offline converation

acidprime:
This was failing for ruby 1.8.7 because of the way `to_s` used to work on arrays , fixed with 0b26d6d675e94b6233e76d28b9357bcf5015a740 if your curious.

PR #236, opened 2015-08-07
Automatically calculate some prefixes

binford2k:
Allows one to set the webhook's `prefix` to:

* `:repo`: the repository name
* `:user`: the owner of the repository
* `:command`: run a command using the posted data to determine prefix
    * This is existing behaviour
    * Uses `prefix command`
    * Aliased to `true` for backwards compatibility
* A string: pass the prefix as a static string
* `false`: disable the prefix

This is currently tested only on GitHub.


acidprime:
:+1: passes acceptance tests

PR #240, opened 2015-08-11
Add single request concurrency option in r10k webhook

pdaukintis:
This ensures that even if webhook is triggered multiple times when r10k is
already running, it would not fail as all requests to Sinatra server are serialised.

By default, this option is set to false. To enable this:

``` Puppet
class {'r10k::webhook::config':
enable_ssl => false,
use_mcollective => false,
enable_mutex_lock => true,
}
```


acidprime:
Specs are failing as we are adding a new default param `enable_mutex_lock` and thus webhook.yaml is not as we think it will be

acidprime:
:+1:

PR #247, opened 2015-09-14
Log when r10k not found in PATH

joshbeard:
This will log at the 'error' level when r10k is not found in the PATH,
thus rendering the agent unusable.  Prior to this, Mcollective logging
would have to be set to 'debug', which would just return that the agent
failed to activate.  This commit hopes to provide more useful
information in this scenario.


acidprime:
@joshbeard thanks, and miss you

Issue #102, opened 2014-11-22
(Feature) webhook should support github secrets


acidprime:
There is  an ability of adding a secret to the payload instead of using basic auth I think the webhook should support his as the password stuff is a little weird in github atm for display and updates.

kenyon:
Looks like this can be closed.

PR #262, opened 2016-01-28
Add GitHub signature

ayohrling:
This addresses issue #102 

- Updated the webhook config to allow specification of the github_secret.
- Added the verify_signature method to the webhook and added the call to verify_signature if the github_secret config value is set.

Validated this working on a FOSS master instance on EL7. Had successful requests with a valid secret and received the 500 error expected when set to an invalid secret value.


acidprime:
Sweet, If you can get the specs updated for the new param I will merge this, otherwise I will get some written for it during the next release and merge it then.

ayohrling:
Added spec test for the additional webhook::config param.

ayohrling:
Went back and merged the latest master into the branch so that the tests can auto-run.

Issue #265, opened 2016-02-19
discussion: r10k deploy environment with no -p argument


mschuchard:
Recently we had a need to update the control repo of a specific directory environment and then update only one module in this environment.  This need arose because we had a special case directory environment with 1200 modules where one module was only ever deployed at a time in a CI/CD pipeline where the mco r10k was executed remotely on a Jenkins build slave.  The -p argument was causing an enormous slowdown on the CD for obvious reasons.

I took our local fork of this module and added a new possible argument (deploy_control_repo) that ostensibly (to the user) omits the -p argument from the mco r10k deploy 'environment' command.  It worked and I was planning on upstreaming the functionality back here.

Stylistically, what would the preference be for me to add this functionality to this module?  There are multiple methods for doing it and I want to ensure the method is ideal before implementing and PR.

igalic:
@mschuchard does https://github.com/acidprime/r10k/pull/266 address your issue?

mschuchard:
Yes.  Let me add this issue to that PR.

rnelson0:
@mschuchard I'm going to close this issue as the PR addresses it. If the PR author doesn't respond shortly we can look at rebasing the PR, or you can if you'd like. Thanks!

PR #266, opened 2016-02-22
Add arguments to allow finer control of deploy behaviour

garethhumphriesig:
- Add "config" argument to allow specifying remote config file for deploy actions
- Add "environment" argument to deploy_module action
- Add "skip_modules" argument to deploy action - just update control repo and skips modules subdir

### DIFF
--- a/templates/agent/r10k.rb.erb
+++ b/templates/agent/r10k.rb.erb
@@ -31,23 +31,31 @@ module MCollective
          'deploy',
          'deploy_module',
          'sync'].each do |act|
+          extra = []
           action act do
+            config = request[:config] || ''
+            if !config.empty?
+              extra += ['-c', config]
+            end
             if act == 'deploy'
               validate :environment, :shellsafe
               environment = request[:environment]
-              run_cmd act, environment
+              skipmodules = request[:skip_modules] || ''
+              if skipmodules.empty?
+                extra += ['-p']
+              end
+              run_cmd act, environment, extra
               reply[:environment] = environment
             elsif act == 'deploy_module'
               validate :module_name, :shellsafe
               module_name = request[:module_name]
               environment = request[:environment] || ''
-              if environment.empty?
-                extra = []
-              else
-                extra = ['-e', environment]
+              if !environment.empty?
+                extra += ['-e', environment]
               end
               run_cmd act, module_name, extra
               reply[:module_name] = module_name
+              reply[:environment] = environment
             else
               run_cmd act
             end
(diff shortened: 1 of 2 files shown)


garethhumphriesig:
- Add config argument to allow specifying config files for deploy actions
 - Add environment argument to deploy_module action
 - Add skip_modules argument to deploy action

mschuchard:
This would close #265.

templates/agent/r10k.rb.erb
```diff
@@ -82,10 +96,11 @@ module MCollective
             cmd << 'cache' if action == 'cache'
             cmd << 'deploy' << 'environment' << '-p' if action == 'synchronize' or action == 'sync'
             if action == 'deploy'
-              cmd << 'deploy' << 'environment' << arg << '-p'
+              cmd << 'deploy' << 'environment' << arg
             elsif action == 'deploy_module'
               cmd << 'deploy' << 'module' << arg
             end
+            cmd += extra
```
mschuchard: Is it safe to be appending this to cache too?

garethhumphriesig: It is in it's present form, as 'extra' is empty in cache context - effectively a no-op.  I put it outside the conditional because it makes extending the other methods the same way in the future easy - anything can alter the way the the r10k call gets executed by just passing in an 'extra' of it's own.

garethhumphriesig:
As as side note - there are surely security implications here, though I don't think anything new.  The assumption is that SSL/authentication/authorisation etc has been done and all actors can be trusted.

rnelson0:
@garethhumphriesig can you rebase this to ensure the changes are still good? Sorry about the delay but we are getting this some attention now!

This will need some documentation as well, please, to ensure users can take advantage of the changes properly.

mschuchard:
Oh wow: I never thought this PR would get revived. This module badly needs this PR. Please get this through.

rnelson0:
@garethhumphriesig bump for update

dhollinger:
@garethhumphriesig another bump for update.

Please see @rnelson0's comments at https://github.com/voxpupuli/puppet-r10k/pull/266#issuecomment-259023489 for what we are waiting for

mschuchard:
I think it is fine to start relaxing the impetus for maintaining a good RPC plugin for r10k since CodeManager is now actually functioning well.

bastelfreak:
Hi. Thanks for raising this PR, but I'm going to close it due to inactivity. Please reopen it if you're still interested in getting this merged.

PR #276, opened 2016-05-20
Add support to overwrite webhook config file permissions

elconas:
Currently webhook.yaml (including the password) is 644 and owned by root. It should be owned by the r10k user and support stricter modes likes 640 or 400.


elconas:
Reopening

PR #279, opened 2016-06-10
unbreak on OS where root group != 'root'

buzzdeee:
Use group -> '0' for /etc/puppetlabs/r10k as well, 
as everywhere else where group 'root' is meant.
This fixes usage on OS that have group 'wheel' as root group, i.e.
OpenBSD.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -49,6 +49,8 @@ class r10k::config (
   $configfile,
   $cachedir,
   $manage_modulepath,
+  $root_user,
+  $root_group,
   $modulepath                = undef,
   $remote                    = '',
   $sources                   = 'UNSET',
@@ -95,16 +97,16 @@ class r10k::config (
   if $configfile == '/etc/puppetlabs/r10k/r10k.yaml' {
     file {'/etc/puppetlabs/r10k':
       ensure => 'directory',
-      owner  => 'root',
-      group  => 'root',
+      owner  => $root_user,
+      group  => $root_group,
       mode   => '0755',
     }
   }
 
   file { 'r10k.yaml':
     ensure  => file,
-    owner   => 'root',
-    group   => '0',
+    owner   => $root_user,
+    group   => $root_group,
     mode    => '0644',
     path    => $configfile,
     content => template($r10k_yaml_template),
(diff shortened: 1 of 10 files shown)


rnelson0:
@buzzdeee it seems like the correct solution would be to make ownership a parameter, with sane defaults for various operating systems, as some people may want to run r10k with different ownership. Is that something you would be willing to work on? If so, be sure to rebase after the most recent changes. Thanks!

buzzdeee:
today was busy day, but I'll see to get around to it a.s.a.p., shouldn't bee too hard to implement.
cheers,

buzzdeee:
what would be a good name for that group? Just 'rootgroup'?
also should the owner of files also be changeable? i.e. wherever root is hardcoded, make that a variable as well?

But then, some next user might start whining, I want to run r10k as user 'r10k' and group 'puppet', and I want configfiles root:puppet 640 and I want 

so, how about having variables like:
$r10k_user -> defines the user running r10k, doing the checkouts/downloads etc.
(IIRC, there is no priviledge dropping implemented in r10k, or is there?, and why should root do some git clones for fetches from puppetforge?) However I think in any case, the, r10k user should be different than the user running the puppet master.

$r10k_group -> the group of the $r10k_user, i.e. group 'puppet' or the like, so the $r10k_user
can share downloaded files with the puppetmaster, but can keep others from looking. i.e. chmod 640. Alternatively, $r10k_user/group could be 'root', and to let puppetmaster run.

Additionally have a $root_user and $root_group.
  - maybe this as first step, replacing all hardcoded user => root and group => root
  - then later on go on with $r10k_user/$r10k_group.

Config files default $root_user:$r10k_group mode 640.
other bin files $root_user:$root_group 0644.

I'd work on it in small steps, starting with:
  - replacing all hardcoded user => root and group => root with respective $root_group, $root_user...

does that sound viable and sane to you?

rnelson0:
@buzzdeee This sounds like a good plan! The default params should be the current values (`root:root` with the exception of whatever you need on BSD) for the time being, even if "best practices" should be separate users. That preserves the current behavior. We can look at changing that in the future.

Thanks!

PR #283, opened 2016-09-22
Ensure webhook run folder exists on redhat systemd environments

luisfdez:
On RHEL7, the `/run` directory is now used as a temporary file storage system (tmpfs), meaning that files there are lost after reboots.

This parameter ensures that the folder is properly created when the service starts.

This implementation relies on systemd functionality that looks appropriate in the context of this module. An alternative, as proposed [here:](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Migration_Planning_Guide/sect-Red_Hat_Enterprise_Linux-Migration_Planning_Guide-File_System_Layout.html#idp4755616), is to define a new configuration file in `/etc/tmpfiles.d/webhook.conf` with contents like:

```
D /var/run/webhook 0755 root root -
```

(Where root can be replace via template with the proper user configured in the module).

### DIFF
--- a/templates/webhook.redhat.service.erb
+++ b/templates/webhook.redhat.service.erb
@@ -5,6 +5,7 @@ After=syslog.target network.target
 [Service]
 Type=forking
 EnvironmentFile=-/etc/sysconfig/webhook
+RuntimeDirectory=webhook
 User=<%= @user %>
 PIDFile=/var/run/webhook/webhook.pid
 TimeoutStartSec=90


rnelson0:
@luisfdez We have fixed the existing issues with tests that caused this to fail originally. Can you rebase this change and fix any failing tests? Thanks.

elconas:
This will always create the directory, even if the service is never started. 

You can add this to /etc/systemd/system/multi-user.target.wants/webhook.service to only create the directories once the service is started.

```
ExecStartPre=/bin/mkdir -p /var/run/webhook
ExecStartPre=/bin/chown r10k /var/run/webhook
PermissionsStartOnly=true
```

elconas:
I think this one solves https://github.com/voxpupuli/puppet-r10k/issues/339

luisfdez:
@rnelson0 sorry for have missed your comment, I just rebased it. It should be fine now.

elconas:
This solves the issue a little different https://github.com/voxpupuli/puppet-r10k/pull/340

dhollinger:
@elconas there is an expectation that if the webhook configuration has been placed but the module, then it should be running. If not, I'd be curious to know why WebBrick isn't properly starting and why I'm not seeing the issue in the puppet log.

PR #281, opened 2016-09-12
Fixes Puppet 4 path for webhook

andrewwippler:
Resolves #273 and Resolves #280. Many thanks to @gitlabgis for finding the path issue for Puppet 4.

PR #285, opened 2016-09-29
On Puppet >= 4 the puppet binaries aren't on the webhook PATH.

gfokkema:
This commit duplicates the check at the top of the file to obtain the path of the puppet binaries.


rnelson0:
@gfokkema could you rebase this now that tests are working? Thanks!

rnelson0:
@gfokkema @andrewwipler actually I think #281 may be the better solution. What do you two think?

rnelson0:
Resolved by #295

PR #288, opened 2016-10-13
Fix testing

ghoneycutt:


rnelson0:
Where is https://travis-ci.org/acidprime/r10k/jobs/167426144#L153-L159 coming from? beaker-rspec and serverspec are in the acceptance group which is skipped.

ghoneycutt:
Yeah, curious why that's getting pulled in

ghoneycutt:
Any ideas on why the acceptance group is not being ignored as it should?

rnelson0:
@ghoneycutt Take a look at #292, I think if you add that to your upudated `.travis.yml`, it will work okay here.

ghoneycutt:
Someone with access should cancel these builds as they will not complete.

rnelson0:
Wow, what the heck happened? I restarted one task and it worked, so I'm restarting the entire build. I think 1.8.7 is still gonna fail, but keeping in mind that we know it will likely fail, I'll still be happy if the system works as we expect it to.

rnelson0:
Yep, it just dies on 1.8.7. I think we will be forced to take that out of the matrix in order to have passing tests.

ghoneycutt:
@rnelson0 I'm good with that. We should make it official and mark the CHANGELOG saying that 1.8.7 support has been dropped, since we are no longer testing with it.

rnelson0:
@ghoneycutt If you want to update the matrix when you get a chance, that would be awesome, but no rush - I think we'll hold off on the merge until the migration to VP is complete. That way we can ensure the `latest` version of `zack/r10k` is a 100% working version and not introduce the chance of problems for anyone not pinning or whose automation grabs the latest.

ghoneycutt:
People who do that want to see the failures or they wouldn't consume it like that :)

rnelson0:
There are a number of tools that grab the latest version off the forge to
automatically update Puppetfiles, and even though 999.999.999 should stand
out as "expired", it would be nice if it was still a working version (this
came up with an accidentally published 999.999.999 version in the wrong
namespace that still managed to collect a non-trivial number of downloads!).

ghoneycutt:
It will work. Though now we are saying that we no longer support ruby 1.8.7 and puppet v2. So if you have those versions, it might work, but that's on you.

ghoneycutt:
Whew! got it. Can I get a merge please?

ghoneycutt:
Anything else left to get this merged?

ghoneycutt:
Thank you!

PR #289, opened 2016-10-13
Drop support for Puppet v2

ghoneycutt:


metadata.json
```diff
@@ -90,11 +90,11 @@
   "requirements": [
     {
       "name": "pe",
-      "version_requirement": "3.x"
+      "version_requirement": ">= 3.1.0 < 4.0.0"
     },
     {
       "name": "puppet",
-      "version_requirement": "2.7.0"
+      "version_requirement": ">= 3.1.0 < 4.0.0"
```
rnelson0: It definitely works with 4.x, how about `>= 3.1.0 < 5.0.0"`?

.travis.yml
```diff
@@ -18,7 +18,6 @@ rvm:
   - 2.0.0
   - jruby-19mode
 env:
-  - PUPPET_VERSION="~> 2.7.0"
```
rnelson0: I think we should test against `~> 3.0` and `~> 4.0`, seems like a good time to make the change if others agree.

It might be worth pairing the RVM/ENV as well because we don't need to test 1.8.7/1.9.3/jruby-19mode with 4.x, for instance.

ghoneycutt:
I was going to do puppet v4 and more ruby versions in separate commits. Hoping to get consensus that we do not need to support puppet v2 anymore.

rnelson0:
Ah, gotcha. You have my vote: No Puppet 2 support.

metadata.json
ghoneycutt: Since we are not testing against Puppet v4, we don't know that. Suggest merging this and then changing to include v4 along with adding it to the test matrix.

rnelson0: Sounds good!

rnelson0:
@ghoneycutt I'm going to close this, we're going to do a modulesync soon and that should bring this up to the proper levels of support.

ghoneycutt:
Cool, thanks!

PR #291, opened 2016-11-02
Update config_version.sh

seanscottking:
I revised your example `config_version.sh` a bit. The major change is that I shortened the result of the SHA code to the first 10 characters. A complete SHA is like 40 characters. I like the cleaner output and thought others would too.

The rest of the changes are simply setting variables at the top of the file for easier configurability/readability

### DIFF
--- a/files/config_version.sh
+++ b/files/config_version.sh
@@ -7,16 +7,16 @@
 # Return the latest commit with the GitHub URL for it
 # The repos are placed using r10k
 
-# Get the hash of the latest commit
-sha=$(git --git-dir \
-  /etc/puppetlabs/puppet/environments/$1/.git rev-parse HEAD)
+ENVIRONMENT="$1"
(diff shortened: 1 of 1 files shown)


ghoneycutt:
Using the full SHA is safer and more accurate.

rnelson0:
I agree with @ghoneycutt.  The rest of the change improves legibility, though.

seanscottking:
Came for the SHA, stay for the ${VAR} 💯

ghoneycutt:
Forgot about #288 .. I need help figuring out why bundler isn't behaving as it should.

seanscottking:
Sorry, I don't know bundler at all

rnelson0:
@seanscottking can you rebate this? It should pass tests now, or at least fail on related changes :)

seanscottking:
OK rebased. Tests are taking a while, let's see if they pass

PR #287, opened 2016-10-13
Duplicate certpath key

seanscottking:
`'certpath'              => $certpath` is defined twice, so I removed the second occurrence of it.

Fixes this message in puppet logs `Warning: The key 'certpath' is declared more than once`

PR #296, opened 2016-11-08
First modulesync after migration

rnelson0:


rnelson0:
It looks like the spec tests need setup with rspec-puppet-facts. I should have time for this tomorrow.

.msync.yml
```diff
@@ -0,0 +1 @@
+modulesync_config_version: '0.15.1'
```
dhoppe: modulesync_config version 0.1.5.1 is not released, yet.

.travis.yml
```diff
@@ -1,65 +1,50 @@
...
-    urls:
-      - 'https://webhooks.gitter.im/e/8936500cc6a369ff46ee'
-    on_success: change
-    on_failure: always
-    on_start: false
+deploy:
+  provider: puppetforge
+  user: puppet
+  password:
+    secure: "ZKiuoKHwvOV4NP0D1iZfxQ2WKQs2QRkoaKviIoT96AjaTuqdYQ7JKr7YNr/AP+V7HeIpwubKbxzJjRpRDLWAB+pOEM7Uguz3hX+KuHjGBlRcfeOfamWj3SFFZomf2LiCXWME1Rv/6qmpCqr49ZZk9ANkSDIvkhuSc/uj0/Tf8o4="
```
dhoppe: This should be set via `.sync.yml`.

manifests/webhook/config.pp
```diff
@@ -53,7 +53,6 @@
       'discovery_timeout'     => $discovery_timeout,
       'certpath'              => $certpath,
       'client_cfg'            => $client_cfg,
-      'certpath'              => $certpath,
```
dhoppe: This has been removed via #287.

spec/spec_helper.rb
```diff
@@ -1,8 +1,31 @@
-require 'rspec-puppet'
+require 'puppetlabs_spec_helper/module_spec_helper'
+require 'rspec-puppet-facts'
+include RspecPuppetFacts
+add_custom_fact :pe_version, '2016.4.2'
```
dhoppe: This should be added to `spec/default_module_facts.yml`.

dhoppe:
@rnelson0 I tried to solve this in my own branch #297. No idea what I should think about the facts and RSpec tests, but at least I am down to 4 errors. Unfortunately I do not have enough time to migrate this to rspec-puppet-facts. :(

rnelson0:
Thanks for the review. I did start this modulesync before the last modulesync config update so I'll definitely get those changes into the correct files.

The spec tests are hurting my brain, hopeful,y someone else can comment on them. I'll keep hacking at it, though.

spec/spec_helper.rb
rnelson0: I think this needs to be added in .sync.yml as a `spec_override` as `default_module_facts.yml` is a flat file that modulesync will override every time.

dhoppe: No. This is a module specific config file an will not be managed via modulesync_config.

bastelfreak: As an example, we do the same here:
https://github.com/voxpupuli/puppet-collectd/blob/master/spec/default_module_facts.yml

rnelson0: Interesting, that file exists in moduleroot of our configs, I assumed that meant it pushed every time. Learn something new every day.

dhoppe: The file `default_facts.yml` is shipped via moduleroot, but we are talking about `default_module_facts.yml` which is a different file name. Just checkout my modulesync branch and play around with modulesync.

manifests/install.pp
```diff
@@ -80,12 +80,10 @@
 
       # Puppet Enterprise 3.8 and ships an embedded r10k so thats all thats supported
       # This conditional should not effect FOSS customers based on the fact
-      unless ($::is_pe == 'true' or $::is_pe == true) and versioncmp("${::pe_version}", '3.8.0') >= 0 and $puppet_master { #lint:ignore:only_variable_string
```
nibalizer: Seems like we lost some conditional logic here

manifests/install/gem.pp
```diff
@@ -3,18 +3,17 @@
   $manage_ruby_dependency,
   $version,
 ) {
-  #include ::r10k
-  require git
+  require ::git
   anchor{'r10k::ruby_done':}
   case $manage_ruby_dependency {
-    include: {
+    'include': {
```
nibalizer: It took me a minute to figure this out, then I became sad

spec/acceptance/prefix_webhook_spec.rb
```diff
@@ -1,9 +1,9 @@
 require 'spec_helper_acceptance'
 require 'json'
 describe 'Prefix Enabled,System Ruby with No SSL, Not protected, No mcollective' do
-
   context 'default parameters' do
-    let(:pp) {"
+    let(:pp) do
+      "
```
nibalizer: Is this really what it wnats? weird

rnelson0: That's what rubocop says is best practice - do instead of { and don't start the block on the same line as the do.

manifests/install.pp
rnelson0: This was one of those "why did it ever pass?" tests. Elsewhere, references were made to `Package[$real_package_name]`, though because of the conditional it might not be present in the catalog, which caused failures. Even though PE 3.8 ships with r10k and the package is already installed, it still needs to be present in the resulting catalog - and this will restore the gem if someone mucks with the puppetserver gem environment.

rnelson0:
Discussed the issues @nibalizer raised on IRC and confirmed that those changes are required and provide no loss of functionality.

dhoppe:
@rnelson0 You are awesome and thx for this great pull request.

PR #299, opened 2016-11-18
Release candidate for v4.0.0

rnelson0:


rnelson0:
It's limited, but I pointed my home server to HEAD and ensured the webhook script had the expected changes and still responded to the webhook event properly.

bastelfreak:
should we migrate the changelog to markdown?

rnelson0:
Tweaked metadata. I think we should convert to github_changelog_generator, at a later date.

PR #300, opened 2016-11-19
rewrite CHANGELOG to markdown

bastelfreak:
our release process requires a `CHANGELOG.md`

```
$ bundle exec rake travis_release
rake aborted!
Errno::ENOENT: No such file or directory @ rb_sysopen - CHANGELOG.md
/home/bastelfreak/modulesync_config/modules/puppet-r10k/.vendor/ruby/2.3.0/bundler/gems/voxpupuli-release-gem-4fc386d54c92/lib/voxpupuli/release/rake_tasks.rb:37:in `readlines'
/home/bastelfreak/modulesync_config/modules/puppet-r10k/.vendor/ruby/2.3.0/bundler/gems/voxpupuli-release-gem-4fc386d54c92/lib/voxpupuli/release/rake_tasks.rb:37:in `block in <top (required)>'
/home/bastelfreak/modulesync_config/modules/puppet-r10k/.vendor/ruby/2.3.0/bundler/gems/voxpupuli-release-gem-4fc386d54c92/lib/voxpupuli/release/rake_tasks.rb:16:in `block in <top (required)>'
/home/bastelfreak/modulesync_config/modules/puppet-r10k/.vendor/ruby/2.3.0/gems/rake-11.3.0/exe/rake:27:in `<top (required)>'
/home/bastelfreak/.gem/ruby/2.3.0/gems/bundler-1.13.3/lib/bundler/cli/exec.rb:74:in `load'
/home/bastelfreak/.gem/ruby/2.3.0/gems/bundler-1.13.3/lib/bundler/cli/exec.rb:74:in `kernel_load'
/home/bastelfreak/.gem/ruby/2.3.0/gems/bundler-1.13.3/lib/bundler/cli/exec.rb:27:in `run'
/home/bastelfreak/.gem/ruby/2.3.0/gems/bundler-1.13.3/lib/bundler/cli.rb:332:in `exec'
/home/bastelfreak/.gem/ruby/2.3.0/gems/bundler-1.13.3/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
/home/bastelfreak/.gem/ruby/2.3.0/gems/bundler-1.13.3/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
/home/bastelfreak/.gem/ruby/2.3.0/gems/bundler-1.13.3/lib/bundler/vendor/thor/lib/thor.rb:359:in `dispatch'
[...]

PR #301, opened 2016-11-20
Rename CHANGELOG.md

rnelson0:


rnelson0:
Supersedes #300.

This will allow us to publish 4.0.0, but is only the start of some pretty significant rewrites. We WILL come back and improve the changelog in the future.

PR #311, opened 2016-12-01
Modulesync v0.16.3

rnelson0:


spec/spec_helper.rb
```diff
@@ -27,4 +29,5 @@
   c.default_facts = default_facts
 end
 
+add_custom_fact :pe_version, '2016.4.2'
```
bastelfreak: can you move this to a `default_module_facts.yml`? https://github.com/voxpupuli/puppet-r10k/blob/master/spec/spec_helper.rb#L26

rnelson0: Remedied, testing is processing slowly of course. Editing from an iPad so can't rebase, the commits will need squashed during merge.

rnelson0:
Thanks!

PR #313, opened 2016-12-03
Modulesync 0.16.4, stop duplicate CI builds

rnelson0:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,9 @@ matrix:
     env: PUPPET_VERSION="~> 4.0" CHECK=test
   allow_failures:
     - rvm: 2.4.0-preview3
+branches:
+  only:
+  - master
 notifications:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -35,6 +35,9 @@ matrix:
     env: PUPPET_VERSION="~> 4.0" CHECK=test
   allow_failures:
     - rvm: 2.4.0-preview3
+branches:
+  only:
+  - master
```
igalic: i think this conflicts with our `deploy` strategy:
```yaml
deploy:
  on:
    tags: true
```

in @brainsware/trafficserver, i added this [hack](https://github.com/Brainsware/puppet-trafficserver/blob/master/.travis.yml#L41):
```yaml
branches:
  only:
  - master
  - /^v\d/
```

Issue #306, opened 2016-11-23
Document breaking changes in CHANGELOG


ghoneycutt:
The CHANGELOG does not contain what the breaking changes are between major versions. This makes it very difficult to upgrade the module.

rnelson0:
@ghoneycutt You can see a sample of a changelog from github_changelog_generator at https://github.com/voxpupuli/puppet-r10k/blob/changelog/CHANGELOG.md. Anything you want to see different in the output to flag this better? You can look at [the options on its wiki](https://github.com/skywinder/github-changelog-generator/wiki/Advanced-change-log-generation-examples) and see if it's possible to do whatever you'd like with some option and I can give it a shot.

ghoneycutt:
Output from the generator is great!

I reported this issue because I went to use v4 and couldn't easily understand from the README or CHANGELOG what the breaking changes are and was unable to easily upgrade.

rnelson0:
That's good to hear as this may be a template for other module changelogs.

WRT r10k, the breaking change has to do with dropping support and tests for Ruby 1.8.7. Though that's a modulesync change that is discarded from the changelog, I think I may have to unlabel that and edit the title to reflect the dropped support, I.e. "BREAKING: Migration to Vox Pupuli drops support and tests for Ruby 1.8.7" or similar. @alexjfisher think that would be a good way to identify that?

rnelson0:
@ghoneycutt I'm going to close this out. Thanks for bringing this up, and please continue to let us know if we do not meet expectations in the future.

PR #314, opened 2016-12-13
Changelog

rnelson0:
Sample to address #306


Gemfile
```diff
@@ -31,6 +31,9 @@ group :test do
   gem 'mocha', '>= 1.2.1',                                          :require => false
   gem 'coveralls',                                                  :require => false if RUBY_VERSION >= '2.0.0'
   gem 'simplecov-console',                                          :require => false if RUBY_VERSION >= '2.0.0'
+  gem 'github_changelog_generator', '~> 1.13.0',                    :require => false if RUBY_VERSION < '2.2.2'
```
ghoneycutt: If these dependencies get crazy you could just note that the person doing the release needs to have a specific ruby version, such as the newest supported.

rnelson0:
This was only a test, closing.

PR #331, opened 2017-01-20
Added Slack webhook

jamtur01:
This adds a configurable Slack notification to the r10k webhook.

See the README for configuration.

### DIFF
--- a/templates/webhook.bin.erb
+++ b/templates/webhook.bin.erb
@@ -57,6 +57,10 @@ if $config['use_mcollective'] then
   include MCollective::RPC
 end
 
+if $config['slack_webhook'] then
+  require 'slack-notifier'
+end
+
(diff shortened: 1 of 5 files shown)


templates/webhook.bin.erb
```diff
@@ -213,11 +264,15 @@ end
         end
         message = run_command(command)
         $logger.info("message: #{message} module_name: #{module_name}")
-        {:status => :success, :message => message.to_s }.to_json
+        status_message = {:status => :success, :message => message.to_s, :module_name => module_name, :status_code => 200}
+        status_message.to_json
```
rnelson0: Should status_message.to_json be the returned value from these methods?

jamtur01: Doh. I didn't even notice. Moved this to the end of the method.

rnelson0:
Thanks!

PR #335, opened 2017-02-11
Modulesync 0.16.11 & Release 4.2.0

bastelfreak:
Hi @rnelson0, I would like to apply or migration thingy to this module:
https://voxpupuli.org/blog/2017/01/11/migrating-to-puppet4/

so modulesync + release 4.2.0. Then sync again. This will be the last release with puppet3 support. is that okay for you?


bastelfreak:
I will hold of with the release until @rnelson0 responds

rnelson0:
I'm definitely okay with the last puppet 3 release. Going to 4 only will be
a bit of a monster but it starts here.
-- 
Rob Nelson

PR #337, opened 2017-02-12
fix webhook systemd init script handling

vchepkov:
This commit fixes several issues:
- systemd unit files should not have executable bit set;
- systemd has to be notified when unit file changes;
- service needs to be restarted, when init script changes ( user id, for example)
- if webhook_bin wasn't installed, webhook_init_script shouldn't either

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -154,15 +154,16 @@ class r10k::params
   $webhook_webrick_version       = '1.3.1'       # Webrick 1.4 requires ruby >= 2.3
   $webhook_generate_types        = false
 
+  $webhook_service_provider      = $::service_provider
   # Service Settings for SystemD in EL7
   if $::osfamily == 'RedHat' and $::operatingsystemmajrelease == '7' {
-    $webhook_service_file     = '/usr/lib/systemd/system/webhook.service'
+    $webhook_service_file     = undef
     $webhook_service_template = 'webhook.redhat.service.erb'
   } elsif $::osfamily == 'Gentoo' {
     $webhook_service_file     = '/etc/init.d/webhook'
     $webhook_service_template = 'webhook.init.gentoo.erb'
   } elsif $::osfamily == 'Suse' and $::operatingsystemrelease >= '12' { #lint:ignore:version_comparison
-    $webhook_service_file     = '/etc/systemd/system/webhook.service'
+    $webhook_service_file     = undef
     $webhook_service_template = 'webhook.suse.service.erb'
   } else {
     $webhook_service_file     = '/etc/init.d/webhook'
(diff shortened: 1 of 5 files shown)


manifests/webhook.pp
```diff
@@ -61,11 +61,31 @@
     before => File['webhook_init_script'],
   }
 
+  if $service_file =~ /systemd/ {
+
+    $_service_file_mode = '0644'
+
+    exec { 'webhook_systemd_reload':
```
bastelfreak: Hi. we try to use https://github.com/camptocamp/puppet-systemd on all Vox Pupuli modules. puppet-systemd provides an exec that can be notified, please use that instead of defining a new one.

vchepkov: I don't mind, but referred module has a design flaw, and was addressed for more then a year, as explained here https://github.com/camptocamp/puppet-systemd/issues/5
Since authors are leaning to change class to 'define', it will be essentially the same implementation

vchepkov: I take it preference is to use systemd module?
I will make a change

bastelfreak: yeah, the implementation may be the same. But I would like to avoid declaring a resource twice, even if it has different names and it would work.

vchepkov:
Took me a while to get back to it, sorry.

vchepkov:
I am not sure why it complains about `$::path` variable.
This should be a standard OS fact, I believe, so I was hesitant to add it to default_facts.yml

manifests/params.pp
```diff
@@ -156,16 +156,20 @@
 
   # Service Settings for SystemD in EL7
   if $::osfamily == 'RedHat' and $::operatingsystemmajrelease == '7' {
-    $webhook_service_file     = '/usr/lib/systemd/system/webhook.service'
+    $webhook_service_file     = undef
```
bastelfreak: instead of defining this for each OS, can we just use the service_provider fact from stdlib?

vchepkov: The idea is let 'systemd' module handle the part you selected

As for the default service provider, sure, I can do that

vchepkov:
Well, now tests fail because two facts are missing.
I would need help to fix those

bastelfreak:
I made a similar fix here: https://github.com/voxpupuli/puppet-prometheus/commit/95bdf8488a476281e9968cf0f5abe4247fb827c4

Does that help you? Otherwise feel fre to join ourc IRC channal #voxpupuli on freenode. We're happy to help out.

vchepkov:
Doesn't seem right to have to populate system facts in each module manually. No generic helper available? I can copy/paste your system_provider fact, but I have no idea what `$::path` should be on those. I will ping you tomorrow. Thank you.

bastelfreak:
the service_provider fact sounds like a core fact, but it's not :( I had the idea to include it in our modulesnc_config, so it's a generic helpe that's always available. I had no time yet to check the logs for the $PATH fact.

Issue #339, opened 2017-02-23
webhook fails to auto-start on Systemd based systems


elconas:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: any
- Ruby: -
- Distribution: Centos7
- Module version: 5.0.0-rc0

## How to reproduce (e.g Puppet code you use)

* Enable webhook with manifest
* reboot system
* test of webhook automatically starts

## What are you seeing

Webhook failes with error message: 

```
Feb 23 12:05:59 SERVER systemd[1]: Starting R10K Webhook Service...
Feb 23 12:06:02 SERVER systemd[1]: PID file /var/run/webhook/webhook.pid not readable (yet?) after start.
Feb 23 12:06:02 SERVER systemd[1]: webhook.service never wrote its PID file. Failing.
Feb 23 12:06:02 SERVER systemd[1]: Failed to start R10K Webhook Service.
Feb 23 12:06:02 SERVER systemd[1]: Unit webhook.service entered failed state.
Feb 23 12:06:02 SERVER systemd[1]: webhook.service failed.
```

## What behaviour did you expect instead

Webhook should start

## Any additional information you'd like to impart

Reason is that systemd cleans up /var/run directory on each boot. This means the /var/run/webhook directory created by the manifest is removed and as the webhook is running as normal user, the directory can not be created.

Possible Fix: 

- Create the directory in Systemd config file
- Use tmpfiles.d to create the directory

rnelson0:
This should be resolved by #283 which was merged about a week after this issue was opened. We will be releasing 5.0.0 soon that will contain the fix.

PR #340, opened 2017-02-23
Fix https://github.com/voxpupuli/puppet-r10k/issues/339

elconas:
Fixes issue https://github.com/voxpupuli/puppet-r10k/issues/339


rnelson0:
The fix in #283 appears to be the more modern resolution. See https://petersouter.co.uk/dealing-with-var-run-in-systemd-unit-files/. Thanks for the PR though!

PR #341, opened 2017-03-03
replace all validate functions with datatypes

bastelfreak:
I had to bump stdlib to 4.13.1 to get the types from it


bastelfreak:
the goal of this is to get rid of the validate functions to get clean puppetserver logs. Its not the goal to set datatypes for all parameter.

Issue #342, opened 2017-03-05
r10k.yaml placement is incorrect


rnelson0:
Starting with Puppet 4, the r10k.yaml placement location for PE is incorrect. At https://github.com/voxpupuli/puppet-r10k/blob/master/manifests/params.pp#L65 it defaults to /etc/r10k.yaml. This generates the warning:

```
WARN     -> The r10k configuration file at /etc/r10k.yaml is deprecated.
WARN     -> Please move your r10k configuration to /etc/puppetlabs/r10k/r10k.yaml.
```

PR #345, opened 2017-03-07
fix for issue: #342 - r10k.yaml placement is incorrect

Kotty666:


rnelson0:
That wasn't the placement I meant, but it looks like a good fix, too!

PR #349, opened 2017-03-12
Webhook Background mode

luckyraul:
option for docker


rnelson0:
@luckyraul Can you describe the difference between these two modes, why a user would want to use one or the other? I'm personally curious, but that should also go somewhere in the readme.

luckyraul:
@rnelson0 

To run webhook as the main process in Docker

rnelson0:
@luckyraul OK, can you add something to the readme then so users are aware of the option and use case, please?

luckyraul:
@rnelson0  check please

luckyraul:
My text is "If you are building your image with the puppet"

rnelson0:
What does that mean? Are you talking about the [image_build module](https://forge.puppet.com/puppetlabs/image_build)?

luckyraul:
It can be that module, or just puppet apply inside the docker's RUN.

Issue #351, opened 2017-03-18
cachedir setting is not idempotent


vchepkov:
'puppet apply' and 'puppet agent' produce different results for the same code:

```
 Notice: /Stage[main]/R10k::Config/File[r10k.yaml]/content: 
--- /etc/puppetlabs/r10k/r10k.yaml	2017-03-18 18:42:23.732745000 +0000
+++ /tmp/puppet-file20170318-18579-18zu7r3	2017-03-18 18:46:56.205745000 +0000
@@ -1,5 +1,5 @@
 ---
-:cachedir: /opt/puppetlabs/puppet/cache/r10k
+:cachedir: /opt/puppetlabs/server/data/puppetserver/r10k
 :sources:
   puppet:
     basedir: /etc/puppetlabs/code/environments
```

vchepkov:
This could be a quirk of puppetserver and the fact that it doesn't honor settings of vardir in puppet.conf and instead using master-var-dir from puppetserver.conf, therefore $settings::vardir are different.

rnelson0:
I think that's probably the case. Can you confirm which of cachedir belongs to which command?

vchepkov:
The top one is for puppet apply, bottom for puppetserver
I see this changed back and forward when PE installer is running, since it's using 'puppet apply' and then reverts it when puppetserver compiles catalog.

vchepkov:
puppetlabs-stdlib provides a fact - $::puppet_vardir, it can be used instead of $settings::vardir, IMHO

PR #352, opened 2017-03-18
use stdlib puppet facts for configuration

vchepkov:
puppetserver doesn't honor puppet.conf vardir settings

This would solve #351


rnelson0:
This broke all the tests. Can you review the failures to see what the fix is?

vchepkov:
As far as I can tell, all failures have nothing to do with my change, seems Travis environment is broken to begin with.

rnelson0:
The errors are definitely related to your changes, they have to do with r10k_basedir. It's expecting an actual path and getting something else. It may be a missing fact or it may need to change the data type. I suggest rebasing on master first, though, there has been a modulesync since this was submitted.

vchepkov:
I will be first one to admin I am bad with tests. I presume the problem is, actual puppet is not running there, so facts are not being collected. Do I add them to default_module_facts.yml  ?

vchepkov:
Well, that didn't help :(

rnelson0:
I think you need to set the facts hash element directly, rspec-puppet doesn't normalize between global facts and the facts hash.

vchepkov:
How do I do this?

vchepkov:
I found something in spec/spec_helper.rb, I'll try to add it there

vchepkov:
I found this PR https://github.com/voxpupuli/puppet-selinux/pull/217/files
And it is using exactly the same changes, but their test succeeds.
I am not sure what I am doing wrong here

rnelson0:
@vchepkov Can you rebase on master? We updated with a modulesync recently, at the least that should make the test results easier to read.

vchepkov:
Done

default_module_facts.yml
```diff
@@ -1,2 +1,4 @@
 ---
 pe_version: '2016.4.1'
+puppet_environmentpath: '/etc/puppetlabs/code/environments'
+puppet_vardir: '/opt/puppetlabs/puppet/cache'
```
rnelson0: Shouldn't this file go in `spec`? That's where [selinux](https://github.com/oranenj/puppet-selinux/blob/36a6b4f1ed540595d67257a972032a605c6321fa/spec/default_module_facts.yml) puts it.

vchepkov: I honestly don't know, I added to an existing file, I assumed it's in the right place
let me move it

vchepkov:
Much better now

rnelson0:
You can blame me for putting the file in the wrong spot :) Thanks!

PR #353, opened 2017-03-20
Use styleguide compliant syntax for sskey example.

kallies:
Replace quotes with single quotes for strings.


bastelfreak:
thanks @kallies!

PR #354, opened 2017-03-21
Webhook - deal with prefix

luckyraul:


rnelson0:
What is the intention with this patch? I'm not familiar with the underlying need, I don't use the prefix parameter.

luckyraul:
puppet-hiera repository does not match hiera source in yaml

bastelfreak:
I think this will break the functionality for everything except for your usecase? I'm fine with this change if it is configureable and the default is the current behavior and not the new one.

luckyraul:
I don't think it can break

```
2.3.3 :001 > 'hiera'.sub(/^.*-/, '')
"hiera"
2.3.3 :002 > 'puppet-hiera'.sub(/^.*-/, '')
"hiera"
```

anyway this code should be very close to 

https://github.com/voxpupuli/puppet-r10k/blob/master/templates/webhook.bin.erb#L98

luckyraul:
This file is not compatible with bitbucket and it loads /etc/r10k.yaml hardcoded 

https://github.com/voxpupuli/puppet-r10k/blob/master/files/prefix_command.rb

rnelson0:
Can you raise the hardcoded path as a separate issue please?
>
>

luckyraul:
This is all about the prefix usage. That file is used for matching json from hook with r10k.yaml's sources.
I don't know the actual reason for such a split. Webhook itself can parse and match.

luckyraul:
added support for bitbucket

luckyraul:
I've spent some time thinking about my change, It is not working correctly, so I created new change to support my situation.

rnelson0:
@luckyraul You said you created a new change to support your situation. Should this be closed then?

luckyraul:
No, this pull request for the same problem. Not working prefix

luckyraul:
@rnelson0 check updated pull request

rnelson0:
@luckyraul I apologize, but I still do not understand the problem you are having or how this addresses it, to properly review the change. It definitely needs some sort of test, as well, but let's start with the problem statement first.

luckyraul:
@rnelson0 I have few modules with prefix in the config. They are installed in folders name_branch.  I want them to be updated via webhook. This is not working at all without changes

binford2k:
The webhook is being moved into its [own tool](https://github.com/voxpupuli/puppet_webhook). Can you evaluate that, and see if your problem still exists please?

PR #356, opened 2017-04-04
Create sanitize_input helper method for strings sent to run_command method.

xraystyle:
I found that this Sinatra app is vulnerable to shell injection when it runs the commands generated by the :deploy and :deploy_module methods. This patch adds a simple helper method to shell-escape this input, and is called on the strings before they're passed to the above methods. Adds no additional dependencies, Shellwords is part of the Ruby standard library.

### DIFF
--- a/templates/webhook.bin.erb
+++ b/templates/webhook.bin.erb
@@ -17,6 +17,7 @@ require 'json'
 require 'yaml'
 require 'cgi'
 require 'open3'
+require 'shellwords'
 
 WEBHOOK_CONFIG = '/etc/webhook.yaml'
 PIDFILE        = '/var/run/webhook/webhook.pid'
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hey @xraystyle, thanks for the PR! Can you please check the used email address in your commit? It isn't associated with your github account.

xraystyle:
Ah, I haven't set an email address in my git config on my work laptop. Now that I've set it, do I need to re-create the pull request?

bastelfreak:
updating the commit + doing a force push is enough.

xraystyle:
Done, should be good now. Author email is one of the emails associated with my github account.

rnelson0:
@xraystyle Great catch here, that's a pretty big security gap. Thank you!

PR #362, opened 2017-04-08
Webhook: ability to pass extra arguments to mco

nike38rus:
This will be useful for finer mco control e.g. to pass `--no-progress` argument to reduce size of access log

PR #363, opened 2017-04-09
Webhook: ability to pass extra arguments to mco

nike38rus:
Continuation of a #362 

This will be useful for finer mco control e.g. to pass `--no-progress` argument to reduce size of access log

Also I made some changes in authentication events logging and fixed identations


rnelson0:
The change itself looks good, just needs some documentation/examples and an rspec test update.

nike38rus:
@rnelson0 it seems I'm done

manifests/webhook/config.pp
```diff
@@ -44,6 +44,7 @@
   $configfile_symlink         = '/etc/webhook.yaml',
   $enable_mutex_lock          = $r10k::params::webhook_enable_mutex_lock,
   Array $ignore_environments  = $r10k::params::webhook_ignore_environments,
+  $mco_arguments              = $r10k::params::webhook_mco_arguments,
```
rnelson0: This should be a `String`, correct?

nike38rus: Well, if user didn't specified this parameter then default `undef` will be used so I guess it should be `Variant[String, Undef]` ?

rnelson0: I think `Optional[String]` is better.

nike38rus: Indeed!

nike38rus:
@rnelson0 done!

rnelson0:
Thanks for your patience!

Issue #365, opened 2017-05-08
webhook gems aren't installable


alexjfisher:
The latest version of rack isn't installable on puppet's ruby 2.1.9.
Up until now, this wasn't a problem as the sinatra gem got installed first and it had a dependency on `rack ~> 1.5`.  (We may as well remove https://github.com/voxpupuli/puppet-r10k/blob/1d0c344b1fee40b6a6d7771ea9b83a9cf07c26c3/manifests/webhook/package.pp#L14-L20)

Sinatra 2.0.0 got released yesterday.  It depends on the newer rack. `rack ~> 2.0`, so we now get
```
Error: Execution of '/opt/puppetlabs/puppet/bin/gem install --no-rdoc --no-ri sinatra' returned 1: ERROR:  Error installing sinatra:
        rack requires Ruby version >= 2.2.2.
Error: /Stage[main]/R10k::Webhook::Package/Package[sinatra]/ensure: change from absent to present failed: Execution of '/opt/puppetlabs/puppet/bin/gem install --no-rdoc --no-ri sinatra' returned 1: ERROR:  Error installing sinatra:
        rack requires Ruby version >= 2.2.2.
```

Due to the `if !defined`, wrapping the installation of sinatra, this is easy enough to work around in my profile.  We should probably fix the issue in the module though.  Is it ok to change
```puppet
package { 'sinatra':
  ensure   => installed,
  provider => $provider,
  before   => Service['webhook'],
}
```
to
```puppet
package { 'sinatra':
  ensure   => '1.4.8',
  provider => $provider,
  before   => Service['webhook'],
}
```
??

rnelson0:
That should be okay, but let's parameterize it with something that defaults to '1.4.8', in case someone wants to provide a specific lower version number.

Re: L14-20, can that be removed? If that's a dependency, seems like the gem provider will install it automatically.

alexjfisher:
When puppet gets around to 14-20 it has nothing to do since rack is already installed (as a sinatra dependency).  Had this not been the case, the module would have been failing since last June (when rack 2.0 came out).

PR #366, opened 2017-05-08
Fix webhook installation by pinning sinatra gem

### LINKED ISSUE #140: (Feature) Webhook should generate a new SSL certificate rather than reusing pe-internal-peadmin-mcollective-client.pem
Currently, the webhook defaults to using the pe-internal-peadmin-mcollective-client SSL certificate.  It would make more sense to generate an SSL cert specifically for the webhook.

alexjfisher:
Sinatra 2.0.0 was released on the 7th of May 2017.  It requires rack ~> 2.0 and *that* requires ruby 2.2.  Puppet 4 AIO ships with ruby 2.1, so pin sinatra to prevent errors.  I've also removed the code that managed rack installation as this was redundant.

Fixes #140

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,17 +1,17 @@
 fixtures:
   repositories:
     stdlib:
-      repo: "git://github.com/puppetlabs/puppetlabs-stdlib.git"
+      repo: "https://github.com/puppetlabs/puppetlabs-stdlib.git"
       ref:  "4.15.0"
-    ruby:    "git://github.com/puppetlabs/puppetlabs-ruby.git"
-    gcc:     "git://github.com/puppetlabs/puppetlabs-gcc.git"
-    pe_gem:  "git://github.com/puppetlabs/puppetlabs-pe_gem.git"
-    make:    "git://github.com/voxpupuli/puppet-make.git"
-    inifile: "git://github.com/puppetlabs/puppetlabs-inifile.git"
-    vcsrepo: "git://github.com/puppetlabs/puppetlabs-vcsrepo.git"
-    git:     "git://github.com/puppetlabs/puppetlabs-git.git"
+    ruby:    "https://github.com/puppetlabs/puppetlabs-ruby.git"
+    gcc:     "https://github.com/puppetlabs/puppetlabs-gcc.git"
+    pe_gem:  "https://github.com/puppetlabs/puppetlabs-pe_gem.git"
+    make:    "https://github.com/voxpupuli/puppet-make.git"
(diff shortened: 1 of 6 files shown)


.fixtures.yml
```diff
@@ -1,17 +1,17 @@
...
+    ruby:    "https://github.com/puppetlabs/puppetlabs-ruby.git"
+    gcc:     "https://github.com/puppetlabs/puppetlabs-gcc.git"
+    pe_gem:  "https://github.com/puppetlabs/puppetlabs-pe_gem.git"
+    make:    "https://github.com/voxpupuli/puppet-make.git"
+    inifile: "https://github.com/puppetlabs/puppetlabs-inifile.git"
+    vcsrepo: "https://github.com/puppetlabs/puppetlabs-vcsrepo.git"
+    git:     "https://github.com/puppetlabs/puppetlabs-git.git"
     portage:
-      repo: "git://github.com/gentoo/puppet-portage.git"
+      repo: "https://github.com/gentoo/puppet-portage.git"
```
alexjfisher: oops.  I didn't actually mean to commit changes to this file.  I can remove them, but then again, they are very useful to those trying to test behind a firewall with only an http/https proxy available for Internet access.

rnelson0: That's a great catch, I have lost so much time to using `git://` where it doesn't work, but never the other way around.

manifests/webhook/package.pp
```diff
@@ -1,19 +1,13 @@
...
     package { 'sinatra':
-      ensure   => installed,
-      provider => $provider,
-      before   => Service['webhook'],
-    }
-  }
-  if !defined(Package['rack']) {
-    package { 'rack':
-      ensure   => installed,
+      ensure   => $sinatra_version,
```
rnelson0: As `~> 1.0` may not work for all possible providers, though it does work with the default provider, can you add some verbiage about it this param in either this file's header as comments, or in the README?

alexjfisher: Ok.  I can't think of any situation where it wouldn't be a gem provider *and* the package would still be called `sinatra` (and not `ruby-sinatra` or something), so a note in the source should be ok?  I'll update it in the morning.

rnelson0: Right, but `installed` works with all providers, whereas `~> 1.0` doesn't work with many of them. This change could be a breaking change to someone using a non-standard provider, so I think it may be helpful to promote it in the README. I'll look tomorrow to see if there's an appropriate place unless you find one first.

alexjfisher: README updated.

alexjfisher:
Fixes #365

Issue #367, opened 2017-05-08
Support using webhook slack notifier through a proxy


alexjfisher:
For R10 itself, I've been able to override the command prefix. ie
```
r10k_command_prefix => 'umask 0022; http_proxy=http://proxy.example.com:3128'
```
But for the slack notifier gem, you need to add some [`http_options`](https://github.com/stevenosloan/slack-notifier#http-options).  I can submit a PR for this.

PR #368, opened 2017-05-09
Add slack_proxy_url parameter

alexjfisher:
Defaults to `undef`.  When set, slack notifications will be sent via the
http proxy given.


rnelson0:
👍

alexjfisher:
Fixes #367

PR #371, opened 2017-05-09
Release 6.0.0

dhollinger:


dhollinger:
Due to the chance that the `$webhook_sinatra_version` default value could cause breakage for users NOT using the default provider, this will be a major version release.

Release created due to the Sinatra 2.0.0 release breaking users using EL7 + defaults for Sinatra installation

rnelson0:
I think #366 got pinned to #140 wrong by @alexjfisher. Alex, any reason #140 should be closed, or was there another issue you meant to tie it to?

alexjfisher:
Oops, good catch. Not sure where I got #140 from. Should have been #367

alexjfisher:
Or #365 even!! (Should really stop trying to use GitHub on the train from my mobile!!)

alexjfisher:
@dhollinger Sorry, you'll have to regenerate the changelog.  I've closed a couple of other tickets and pinged Rob about another.  Hopefully by the time you come online everything should be 'correct' and ready to go.

PR #385, opened 2017-08-31
Add logging messages

amateo:
Logs when the webhook starts deploying an environment or module


rnelson0:
Thanks, this should really help with visibility in the logs.

PR #389, opened 2017-10-28
Generate types option

dhollinger:
After running into an issue where changes in the `concat_file` type in `puppetlabs-concat` was not being picked up properly and discussing solutions/workarounds with @eputnam, I decided to add functionality to the webhook that allows the execution of `puppet generate types` on an environment after an environment deploy.

This addition adds a new setting in webhook.yaml called `generate_types`. The config option takes a Boolean value and defaults to false. If set to true, then the deploy method will also call the `generate_types` method and pass the control repo's branch name to the method which will be used for the environment parameter when `generate_types` executes the `puppet generate types` command.

For more information on `puppet generate` see https://puppet.com/docs/puppet/4.6/man/generate.html

### DIFF
--- a/templates/webhook.bin.erb
+++ b/templates/webhook.bin.erb
@@ -154,15 +154,6 @@ class Server < Sinatra::Base
         data['repository']['default_branch']                     # github tagged release; no ref.
       ).sub('refs/heads/', '') rescue nil
 
-    deleted = (
-      data['deleted'] || # Github
-      data['after']      # Github and Gitlab
-    )
-
-    if deleted == true || deleted == '0000000000000000000000000000000000000000'
-      branch_deleted = true
-    end
-
     # If prefix is enabled in our config file, determine what the prefix should be
     prefix = case $config['prefix']
     when :repo
@@ -177,7 +168,15 @@ class Server < Sinatra::Base
 
     # When a branch is being deleted, a deploy against it will result in a failure, as it no longer exists.
     # Instead, deploy the default branch, which will purge deleted branches per the user's configuration
-    deleted = data['deleted'] rescue false
+    branch_deleted = (
+      data['deleted'] rescue false || # Github
+      data['after']      # Github and Gitlab
+    )
+
+    if branch_deleted == true || deleted == '0000000000000000000000000000000000000000'
+      deleted = true
+    end
+
     if deleted
       branch = $config['default_branch']
     else
@@ -198,7 +197,7 @@ class Server < Sinatra::Base
       return 200
     else
       $logger.info("Deploying environment #{env}")
-      deploy(env, branch_deleted)
+      deploy(env, deleted)
     end
   end
 
--- a/templates/webhook.bin.erb
+++ b/templates/webhook.bin.erb
@@ -168,9 +168,10 @@ class Server < Sinatra::Base
 
     # When a branch is being deleted, a deploy against it will result in a failure, as it no longer exists.
     # Instead, deploy the default branch, which will purge deleted branches per the user's configuration
-    branch_deleted = (
+    deleted = (
       data['deleted'] rescue false || # Github
-      data['after']      # Github and Gitlab
+      data['push']['changes'][0]['closed'] rescue false || # Bitbucket
+      data['after] rescue false     # Gitlab
     )
 
     if branch_deleted == true || deleted == '0000000000000000000000000000000000000000'
--- a/templates/webhook.bin.erb
+++ b/templates/webhook.bin.erb
@@ -267,13 +267,10 @@ class Server < Sinatra::Base
         $logger.info("message: #{message} environment: #{environment}")
         status_message = {:status => :success, :message => message.to_s, :environment => environment, :status_code => 200}
         notify_slack(status_message) if slack?
-        status_message.to_json
       rescue => e
         $logger.error("message: #{e.message} trace: #{e.backtrace}")
-        status 500
         status_message = {:status => :fail, :message => e.message, :trace => e.backtrace, :environment => environment, :status_code => 500}
         notify_slack(status_message) if slack?
-        status_message.to_json
       end
     end
 
--- a/templates/webhook.bin.erb
+++ b/templates/webhook.bin.erb
@@ -154,6 +154,15 @@ class Server < Sinatra::Base
         data['repository']['default_branch']                     # github tagged release; no ref.
       ).sub('refs/heads/', '') rescue nil
 
+    deleted = (
+      data['deleted'] || # Github
+      data['after']      # Github and Gitlab
+    )
+
+    if deleted == true || deleted == '0000000000000000000000000000000000000000'
+      branch_deleted = true
+    end
+
     # If prefix is enabled in our config file, determine what the prefix should be
     prefix = case $config['prefix']
     when :repo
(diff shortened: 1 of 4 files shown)


alexjfisher:
Excellent! This had been on my todo list for ages. I’ll test it out on Monday.

templates/webhook.bin.erb
```diff
@@ -352,6 +370,7 @@ class Server < Sinatra::Base
         end
         status_message =  {:status => :success, :message => message.to_s, :branch => branch, :status_code => 200}
         $logger.info("message: #{message} branch: #{branch}")
+        generate_types(branch) if types?
```
alexjfisher: What happens if the branch is being deleted?

In general, I've always thought it would be good if somewhere we set whether the hook was for a branch create, modify or delete.  I guess doing this would need a bit of a refactor.  The way we set 'branch' is already a little ugly.  Maybe we could extract a whole bunch of things.  It'd be quite nice to be able to include a bit more detail in the slack notifications etc.

templates/webhook.bin.erb
```diff
@@ -154,6 +154,15 @@ class Server < Sinatra::Base
         data['repository']['default_branch']                     # github tagged release; no ref.
       ).sub('refs/heads/', '') rescue nil
 
+    deleted = (
+      data['deleted'] || # Github
+      data['after']      # Github and Gitlab
+    )
+
+    if deleted == true || deleted == '0000000000000000000000000000000000000000'
+      branch_deleted = true
```
alexjfisher: maybe a `deleted?` method that could be called from `deploy` would be better??

dhollinger: After looking into building a `delete?` I realized that it would be difficult without passing the payload data from the endpoint through the `deploy` method into the `delete?` method (at least based on my limited understanding of Sinatra). 

Definitely seems like something we should do, but may be better to wait for a larger refactor of the code.

alexjfisher: Line 180 `deleted = data['delered'] rescue false` can go now though?

templates/webhook.bin.erb
```diff
@@ -250,6 +259,24 @@ class Server < Sinatra::Base
       message
     end
 
+    def generate_types(environment)
+      begin
+        command = "#{$command_prefix} /opt/puppetlabs/bin/puppet generate types --environment #{environment}"
```
alexjfisher: In my environments, I get a handful of errors, but `generate types` still returns `0`.
eg
```
[root@puppet4server-ci01c ~]# umask 0022;  /opt/puppetlabs/bin/puppet generate types --environment GIT_dev_generate_types
Notice: Generating Puppet resource types.
Error: /etc/puppetlabs/code/environments/GIT_dev_generate_types/modules/openldap/lib/puppet/type/openldap_access.rb: title patterns that use procs are not supported.
Error: Could not autoload puppet/provider/openldap_database/augeas: undefined method `provider' for nil:NilClass
Error: Failed to load custom type 'openldap_database' from '/etc/puppetlabs/code/environments/GIT_dev_generate_types/modules/openldap/lib/puppet/type/openldap_database.rb': Could not autoload puppet/provider/openldap_database/augeas: undefined method `provider' for nil:NilClass
Error: /etc/puppetlabs/code/environments/GIT_dev_generate_types/modules/openldap/lib/puppet/type/openldap_dbindex.rb: title patterns that use procs are not supported.
Error: /etc/puppetlabs/code/environments/GIT_dev_generate_types/modules/openldap/lib/puppet/type/openldap_overlay.rb: title patterns that use procs are not supported.
Error: /etc/puppetlabs/code/environments/GIT_dev_generate_types/modules/wildfly/lib/puppet/type/wildfly_resource.rb: title patterns that use procs are not supported.
Error: /etc/puppetlabs/code/environments/GIT_dev_generate_types/modules/yamlfile/lib/puppet/type/yaml_setting.rb: title patterns that use procs are not supported.
[root@puppet4server-ci01c ~]# echo $?
0
```

templates/webhook.bin.erb
dhollinger: Hmm, not sure how I missed that. We'll actually need some of that. I'll update the PR so that block can use the values from gitlab and bitbucket.

templates/webhook.bin.erb
```diff
@@ -250,6 +259,24 @@ class Server < Sinatra::Base
...
+
+        message = run_command(command)
+        $logger.info("message: #{message} environment: #{environment}")
+        status_message = {:status => :success, :message => message.to_s, :environment => environment, :status_code => 200}
+        notify_slack(status_message) if slack?
+        status_message.to_json
```
alexjfisher: where does this go?  I think in deploy it's the json that ends up being returned to the http client.

dhollinger: Good point, I removed the status_message.to_json line.

I left the rest of the code there. We want the logging to happen still and we want to be able to send notifications to slack if generate_types succeeds or fails.

templates/webhook.bin.erb
```diff
@@ -168,7 +168,16 @@ class Server < Sinatra::Base
 
     # When a branch is being deleted, a deploy against it will result in a failure, as it no longer exists.
     # Instead, deploy the default branch, which will purge deleted branches per the user's configuration
-    deleted = data['deleted'] rescue false
+    deleted = (
+      data['deleted'] rescue false || # Github
+      data['push']['changes'][0]['closed'] rescue false || # Bitbucket
```
alexjfisher: Bitbucket *server* (or stash) needs to look for `data['refChanges'][0]['type'] == 'DELETED'`
See https://confluence.atlassian.com/bitbucketserver/post-service-webhook-for-bitbucket-server-776640367.html

rnelson0:
@alexjfisher How is this working out for you? Any performance impacts we should note in the readme?

alexjfisher:
@rnelson0 It's looking pretty good.  I had some changes to the slack notifications, but I think all our efforts have moved into creating a new gem. https://github.com/voxpupuli/puppet-webhook

Any input you have into that *very* much appreciated.

rnelson0:
I am not as good at Ruby as you may think I am, but I will look :)

rnelson0:
@alexjfisher I know there's the new webhook gem effort, but if this is working for you, we could get it merged now. I assume it will be a bit before the gem is complete and integrated and esp with concat changes this may be helpful sooner rather than later.

alexjfisher:
It was pretty much working. One issue is that puppet generate types returns 0 even if there are errors. I started looking at making improvements, but my changes started getting quite invasive. If @dhollinger is happy, I think this can be merged as is and we can make improvements in the gem version.

dhollinger:
I'm happy with this. If no one else has concerns or objections, I say merge it

templates/webhook.bin.erb
```diff
@@ -168,9 +168,10 @@ class Server < Sinatra::Base
 
     # When a branch is being deleted, a deploy against it will result in a failure, as it no longer exists.
     # Instead, deploy the default branch, which will purge deleted branches per the user's configuration
-    branch_deleted = (
+    deleted = (
       data['deleted'] rescue false || # Github
-      data['after']      # Github and Gitlab
+      data['push']['changes'][0]['closed'] rescue false || # Bitbucket
+      data['after] rescue false     # Gitlab
```
alexjfisher: @dhollinger How did we miss the syntax error on this line??

alexjfisher: Fixed in https://github.com/voxpupuli/puppet-r10k/pull/396

PR #395, opened 2017-11-30
Include environment prefix in Slack notification

spacepants:
When using environment prefixes with r10k, Slack notifications can sometimes get a bit confusing.

This adds the environment prefix to the notification.

### DIFF
--- a/templates/webhook.bin.erb
+++ b/templates/webhook.bin.erb
@@ -308,6 +308,12 @@ class Server < Sinatra::Base
                  http_options: http_options
       end
 
+      if prefix.nil? or prefix.empty?
+        target_prefix = ''
+      else
+        target_prefix = "#{prefix}_"
+      end
+
       if status_message[:branch]
         target = status_message[:branch]
       elsif status_message[:module]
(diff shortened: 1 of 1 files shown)


alexjfisher:
I was pretty sure the slack notification already contained the prefix... ??
@spacepants Which endpoint are you using?  `/payload', or '/module' ?

alexjfisher:
Ok. It's worse than I thought.  This *break* notifications completely if you're using `/payload`

```
ERROR message: undefined local variable or method `prefix' for #<Server:0x007f9e742446d0> trace: ["/usr/local/bin/webhook:302:in `notify_slack'"
```

alexjfisher:
@spacepants Sorry we had to revert this, but we urgently needed to release 6.3.1

If you want to have another go, would you like to create a PR on https://github.com/voxpupuli/puppet_webhook instead?  The webhook is becoming its own gem and soon this module will use that instead.

spacepants:
Whoops. Sorry about that.

Figured out the issue I was seeing on my end. I was using the default `prefix_command.rb` which is looking for r10k's config (or the symlink) at `/etc/r10k.yaml`, but the default config path lives at `/etc/puppetlabs/r10k/r10k.yaml` and `$manage_configfile_symlink ` defaults to `false` which means it wasn't actually parsing the r10k config for the prefix.

I didn't notice because our primary control repo isn't prefixed, so those notifications were coming through just fine. Sorry for the confusion.

PR #400, opened 2017-12-07
Add deploy task

binford2k:
#puppethack This is a pretty basic and not very clever task to orchestrate r10k
deployments. I'd like to add more at some point, but this seems like an
ok MVP.

### DIFF
--- a/tasks/deploy.json
+++ b/tasks/deploy.json
@@ -1,0 +1,22 @@
+{
+  "description": "Trigger an r10k deployment.",
+  "puppet_task_version": 1,
+  "parameters": {
+    "environment": {
+      "description": "The environment to deploy.",
+      "type": "Pattern[/\A[a-z0-9_]+\Z/]"
+    },
+    "environments": {
+      "description": "A comma separated list of environments to deploy. Takes precedence over `environment`.",
+      "type": "Pattern[/\A[a-z0-9_,]+\Z/]"
+    },
+    "module": {
+      "description": "A module to deploy across all environments.",
+      "type": "Pattern[/\A[a-z0-9_]+\Z/]"
+    },
+    "modules": {
+      "description": "A comma separated list of modules to deploy. Takes precedence over `module`.",
+      "type": "Pattern[/\A[a-z0-9_,]+\Z/]"
+    }
+  }
+}
--- a/tasks/deploy.rb
+++ b/tasks/deploy.rb
@@ -1,0 +1,37 @@
+#!/opt/puppetlabs/puppet/bin/ruby
+require 'json'
+require 'open3'
+
+params       = JSON.parse(STDIN.read)
+
+# Munge items into arrays of things to deploy
+environments = params['environments'] || params['environment']
+environments = environments.split(',') if environments
+
+modules      = params['modules'] || params['module']
+modules      = modules.split(',') if modules
+
+messages   = []
+exitstatus = 0
+if environments
+  environments.each do |env|
+    output, status = Open3.capture2e('r10k', 'deploy', 'environment', env, '--puppetfile')
+    messages << "Deploying environment #{env}: #{output}"
+    exitstatus += status.exitstatus
+  end
+end
+
+if modules
+  output, status = Open3.capture2e('r10k', 'deploy', 'module', *modules)
+  messages << "Deploying modules #{modules.join(',')}: #{output}"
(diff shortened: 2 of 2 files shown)


tasks/deploy.rb
```diff
@@ -0,0 +1,37 @@
...
+    messages << "Deploying environment #{env}."
+    `/opt/puppetlabs/bin/r10k deploy environment #{env} --puppetfile`
+    exitstatus += $?.exitstatus
+  end
+end
+
+if modules
+  list = modules.join(',')
+  messages << "Deploying modules #{list}."
+  `/opt/puppetlabs/bin/r10k deploy module #{list}`
```
bastelfreak: can we rely on a proper PATH variable here? Otherwise we need a solution to set the path or something similar. This will fail on systems where r10k isn't installed into puppet.

bastelfreak:
Puppet tasks \o/

tasks/deploy.json
```diff
@@ -0,0 +1,22 @@
+{
+  "description": "Trigger an r10k deployment.",
+  "puppet_task_version": 1,
+  "parameters": {
+    "environment": {
```
binford2k: It kind of seems stupid to have singular and plural forms of these arguments.

tasks/deploy.json
```diff
@@ -0,0 +1,22 @@
...
+      "description": "The environment to deploy.",
+      "type": "String"
```
adreyer: Since there are simple regexes for module and environment names we should probably use them to validate either in the metadata types or the task script itself to prevent shell injections.

Since there are two options here these should probably be optional. 

something like
 Optional[Pattern[/\A[a-z0-9_]+\Z/]]

tasks/deploy.rb
```diff
@@ -0,0 +1,37 @@
...
+
+modules      = params['modules'] || params['module']
+modules      = modules.split(',') if modules
+
+messages   = []
+exitstatus = 0
+if environments
+  environments.each do |env|
+    messages << "Deploying environment #{env}."
+    `/opt/puppetlabs/bin/r10k deploy environment #{env} --puppetfile`
```
adreyer: My recollection of r10k is that it will be hard to debug failures without capturing stderr. I could be wrong though

binford2k: r10k is hard to debug anyway. My thought was that if it failed, you'd attempt a manual deploy for troubleshooting.

tasks/deploy.json
rnelson0: If you're going to do that, I am pretty sure you cannot start or end with an underscore, not sure about leading digits. I don't know where the official restriction description/regex is located, though.

tasks/deploy.rb
rnelson0: It might be worthwhile to allow an optional path parameter, for those who do things their own way.

tasks/deploy.rb
binford2k: I don't have a solid answer for that, but relying on `$PATH` seemed to work for me, so I took out the fully qualified path.

bastelfreak:
:+1: for the last commit message

tasks/deploy.rb
```diff
@@ -0,0 +1,37 @@
...
+  exitstatus += status.exitstatus
+end
+
+if environments.nil? && modules.nil?
+  output, status = Open3.capture2e('r10k', 'deploy', 'environment', '--puppetfile')
+  messages << "Deploying all environments: #{output}"
+  exitstatus += status.exitstatus
+end
+
+puts({ 'messages' => messages }.to_json)
```
rnelson0: Should the exit status be returned in the json as well?

binford2k: hmm. There's not a precedent for it, but I do think that's probably a good idea. Maybe even a more structured output instead of just a list of messages and a status. Lemme think on it.

tasks/deploy.rb
```diff
@@ -0,0 +1,55 @@
...
+  unless output.empty?
+    (status.success? ? warnings : errors) << output
+  end
+end
+
+puts({
+  'status'   => exitstatus.zero?,
+  'messages' => messages,
+  'warnings' => warnings,
+  'errors'   => errors
```
binford2k: I hate rubocop and it's wrong. https://github.com/bbatsov/ruby-style-guide#no-trailing-array-commas

PR #401, opened 2017-12-19
Pin webrick gem

D4rkSh1t:
Webrick 1.4.1 was released on the 18th of becember 2017. It requires ruby >= 2.3.0.

```
Error: /Stage[main]/R10k::Webhook::Package/Package[webrick]/ensure: change
from absent to present failed: Execution of
'/opt/puppetlabs/puppet/bin/gem install --no-rdoc --no-ri webrick'
returned 1: ERROR:  Error installing webrick: webrick requires Ruby
version >= 2.3.0
```
And the provider 'puppet_gem' has only with ruby version 2.2 (Puppet 4)

### DIFF
--- a/README.md
+++ b/README.md
@@ -493,7 +493,7 @@ If you are overriding `r10k::webhook::package::provider`, you will also need to 
 
 By default, the `r10k::webhook::package` class uses the `puppet_gem` provider to install the webrick gem (version 1.3.1)
 This version is compatible with all ruby version (webrick version 1.4.1 require Ruby version >=2.3.0)
-If you are overriding `r10k::webhook::package::provider`, you will also need to override `r10k::webhook::package::webrick_version`.
+If you are overriding `r10k::webhook::package::provider`, you may also override `r10k::webhook::package::webrick_version`.
 
 ### Webhook Slack notifications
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -489,6 +489,12 @@ class { 'r10k::webhook':
 By default, the `r10k::webhook::package` class uses the `puppet_gem` provider to install the latest ruby 2.1 compatible version of sinatra ('~> 1.0').
 If you are overriding `r10k::webhook::package::provider`, you will also need to override `r10k::webhook::package::sinatra_version`.
 
+### Webhook webrick gem installation
+
+By default, the `r10k::webhook::package` class uses the `puppet_gem` provider to install the webrick gem (version 1.3.1)
+This version is compatible with all ruby version (webrick version 1.4.1 require Ruby version >=2.3.0)
+If you are overriding `r10k::webhook::package::provider`, you will also need to override `r10k::webhook::package::webrick_version`.
```
rnelson0: will, or may?

D4rkSh1t: may is right. 
Should I change it?

rnelson0: Yes, please!

dhollinger: Should we pin webrick to 1.3.1? webrick <= 1.3.1 has some serious security issues (github reminds me daily about it)

rnelson0: 1) I think we definitely need the parameter, which this PR brings
2) We could then extend the logic to have better defaults per OS if we'd like, still leaving room for overrides

I think # 1 is pretty important as this simply fails right now on new installs, and if we can get #2 in as well, great, but #2 can be addressed by specifying an override.

dhollinger: I think having the parameter with a default being `1.3.1` should be ok for now. The affected ruby versions in the WEBrick CVE aren't supported by voxpupuli anyway.

PR #403, opened 2018-01-07
Bump upper limit of the make dependency

toepi:


bastelfreak:
Thanks for the PR @toepi !

PR #405, opened 2018-01-13
Add LSB tags to the webhook init script.

dickp:
This fixes the error returned when trying to run 'systemctl enable webhook' on
Ubuntu 16.04:

insserv: warning: script 'K01webhook' missing LSB tags and overrides
insserv: warning: script 'webhook' missing LSB tags and overrides
update-rc.d: error: webhook Default-Start contains no runlevels, aborting.

### DIFF
--- a/templates/webhook.init.erb
+++ b/templates/webhook.init.erb
@@ -7,6 +7,15 @@
 #
 # description: Enables web based runs of r10k i.e. github enterprise post hooks
 # processname: webhook
+#
+### BEGIN INIT INFO
+# Required-Start:    $remote_fs $network
+# Required-Stop:     $remote_fs $network
+# Provides:          webhook
+# Default-Start:     2 3 4 5
+# Default-Stop:      0 1 6
+# Short-Description: Init script for running the r10k webhook daemon
+### END INIT INFO
 
 [ -f /etc/sysconfig/webhook ] && . /etc/sysconfig/webhook
 


templates/webhook.init.erb
```diff
@@ -7,6 +7,15 @@
 #
 # description: Enables web based runs of r10k i.e. github enterprise post hooks
 # processname: webhook
+#
+### BEGIN INIT INFO
+# Required-Start:    $remote_fs $network
+# Required-Stop:     $remote_fs $network
```
rnelson0: I don’t think this is right. Were these parameters copied from another script?

dickp: Do you mean the $remote_fs part?

To answer your question, yes I took inspiration from the other examples in /etc/init.d on my system.  Almost every script that depended on $network also depended on $remote_fs, and I thought it not impossible that webhook would also be used with a network filesystem.

rnelson0: Oh, I did not realize those were variables for systemd. Carry on.

rnelson0:
@dhollinger Do you want to take a look at this? I do not know systemd init scripts that well, or how this dovetails into the planned use of the webhook gem.

templates/webhook.init.erb
dhollinger: LSB tags only apply to init scripts, not systemd. (Took some time to find that out).

Unless Debian/Ubuntu are doing something with systemd that no one else is, this is only applicable to the sysVInit scripts.

That being said, this appears to be the template for the init script, not the systemd unit file, so I have no issues with this change. Especially since it's primarily a documentation/non-functional and non-breaking change

PR #406, opened 2018-01-23
Implement locking for mcollective r10k to avoid multiple instances running in parallel

treydock:
We've had several instances where one person runs r10k near same time as another using mcollective which sometimes results in entire environments getting purged and breaking puppet.  r10k itself doesn't yet support locking, https://tickets.puppetlabs.com/browse/RK-125, so implementing the locking in the mcollective agent.

This change in a test environment:

```
$ mco r10k synchronize

 * [ ============================================================> ] 1 / 1

puppet-test.DOMAIN:
Another instance of r10k already running

Finished processing 1 / 1 hosts in 10.64 ms

```


rnelson0:
@treydock thanks for these mcollective-related PRs!

PR #410, opened 2018-02-22
Using #to_yaml instead of printing each element

danzilio:
This PR modifies the template for `r10k.yaml` to use the `#to_yaml`
method to generate valid YAML data structures from the data passed to
the template. Previously, we'd print each element assuming there would
be no nested data structures. This PR allows nested data structures to
be passed to `git_settings`, `forge_settings`, and `deploy_settings`
parameters.


bastelfreak:
Thanks for the PR @danzilio !

PR #413, opened 2018-03-16
Add supports for rocketchat

amateo:
In the same way that webhook could send updates to slack, this commit
adds support to send updates to rocketchat.


rnelson0:
Awesome, thanks!

Issue #412, opened 2018-03-12
generate types feature doesn't work under unprivileged user id


vchepkov:
We run webhook service under 'puppet' user id, puppet generate types fails, since environment is modified for non-root users 

From /var/log/webhook/access.log

```
[2018-03-12 13:39:11] ERROR message: \n/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/environments.rb:38:in `get!': Could not find a directory environment named 'hook_test' anywhere in the path: /opt/puppetlabs/server/data/puppetserver/.puppetlabs/etc/code/environments. Does the directory exist?
```

puppet-r10k (v6.3.2)

vchepkov:
```
# getent passwd puppet
puppet:x:997:995:puppetserver daemon:/opt/puppetlabs/server/data/puppetserver:/sbin/nologin
```

Webhook needs to pass additional arguments for generate types:

```
# sudo -u puppet /opt/puppetlabs/bin/puppet generate types --environment hook_test --confdir /etc/puppetlabs/puppet --codedir /etc/puppetlabs/code --vardir /opt/puppetlabs/puppet/cache
Notice: Generating Puppet resource types.
Warning: The a2mod provider is deprecated, please use apache::mod instead
Notice: Generating '/etc/puppetlabs/code/environments/hook_test/.resource_types/a2mod.pp' using 'pcore' format.
Notice: Generating '/etc/puppetlabs/code/environments/hook_test/.resource_types/augeasprovider.pp' using 'pcore' format.
Notice: Generating '/etc/puppetlabs/code/environments/hook_test/.resource_types/grub_config.pp' using 'pcore' format.
Notice: Generating '/etc/puppetlabs/code/environments/hook_test/.resource_types/grub_menuentry.pp' using 'pcore' format.
Notice: Generating '/etc/puppetlabs/code/environments/hook_test/.resource_types/grub_user.pp' using 'pcore' format.
Notice: Generating '/etc/puppetlabs/code/environments/hook_test/.resource_types/kernel_parameter.pp' using 'pcore' format.
Notice: Generating '/etc/puppetlabs/code/environments/hook_test/.resource_types/sysctl.pp' using 'pcore' format.
Notice: Generating '/etc/puppetlabs/code/environments/hook_test/.resource_types/concat_file.pp' using 'pcore' format.
[...]

PR #414, opened 2018-03-18
(#412) explicitly set working directories to work for non-root accounts

vchepkov:

### DIFF
--- a/templates/webhook.bin.erb
+++ b/templates/webhook.bin.erb
@@ -264,7 +264,13 @@ class Server < Sinatra::Base
 
     def generate_types(environment)
       begin
-        command = "#{$command_prefix} /opt/puppetlabs/bin/puppet generate types --environment #{environment}"
+<% if @user != 'root' -%>
+        # explicitly set working directories to work for non-root accounts
+        generate_suffix = '--confdir /etc/puppetlabs/puppet --codedir /etc/puppetlabs/code --vardir /opt/puppetlabs/puppet/cache'
+<% else -%>
+        generate_suffix = ''
+<% end -%>       
+        command = "#{$command_prefix} /opt/puppetlabs/bin/puppet generate types --environment #{environment} #{generate_suffix}"
 
         message = run_command(command)
         $logger.info("message: #{message} environment: #{environment}")


templates/webhook.bin.erb
```diff
@@ -264,7 +264,13 @@ class Server < Sinatra::Base
 
     def generate_types(environment)
       begin
-        command = "#{$command_prefix} /opt/puppetlabs/bin/puppet generate types --environment #{environment}"
+<% if @user != 'root' -%>
+        # explicitly set working directories to work for non-root accounts
+        generate_suffix = '--confdir /etc/puppetlabs/puppet --codedir /etc/puppetlabs/code --vardir /opt/puppetlabs/puppet/cache'
```
rnelson0: If the user is non-root, it seems possible they could use different paths. Should these be configurable, using those paths as the default?

vchepkov: They definitely use different paths, that's what I am trying to fix.

Those defaults is where typical paths are. I guess we can pass them as arguments, but ideally we should determine them from 'settings' hash.

rnelson0: Either way seems better than static content that may be incorrect.

vchepkov: I agree. What method do you prefer?
- add additional parameters to the class
- add facts providing current configuration

I think former is more flexible, in case webhook gains functionality to execute generate types on a remote master

rnelson0: Hrm. I'm wondering if maybe this is a stopgap anyway and should go in the webhook's gem. @dhollinger thoughts?

dhollinger: This has not been accounted for in the webhook gem yet. Probably create the stopgap here and then open an issue on the gem repo.

rnelson0: @vchepkov sorry for the delay. Since I think you're the primary user of this in the short term, are you happy with this stopgap and then will you open an issue on [webhook](https://github.com/voxpupuli/puppet_webhook)?

vchepkov:
I suspect not many people aware of generate types yet, which is default in PE for awhile. I will open ticket in webhook repository shortly

PR #415, opened 2018-03-22
Fix metadata for the deploy task

dnlsng:
Currently, this file does not contain valid JSON as the backslashes need to be escaped.


bastelfreak:
thanks @dnlsng !

PR #419, opened 2018-03-29
Restart webhook service after package update/installation

tuxmea:
When upgrading webhook gems or when switching to newer puppet with newer
ruby, the webhook service must be restarted.


rnelson0:
Thanks, I hate when this happens and I'm stuck wondering why my changes aren't making it to the server!

PR #435, opened 2018-06-03
remove ::git dependency from r10k::install::puppet_gem

Andor:

### DIFF
--- a/manifests/install/puppet_gem.pp
+++ b/manifests/install/puppet_gem.pp
@@ -1,6 +1,5 @@
 # This class links the r10k binary for Puppet FOSS 4.2 and up
 class r10k::install::puppet_gem {
-  require ::git
 
   if versioncmp("${::puppetversion}", '4.2.0') >= 0 { #lint:ignore:only_variable_string
     file { '/usr/bin/r10k':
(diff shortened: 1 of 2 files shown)


bastelfreak:
@rnelson0 any ideas if we can safely remove the dependency? I think it shouldn't be an issue?

rnelson0:
While the puppet code does not need git, we [expect a git binary](https://github.com/voxpupuli/puppet-r10k/blob/99326494e3b655548ff99ca2a109e8d5d6ff59e2/files/config_version.sh#L14) to be there. Which is what the git module does. I think it should be left.

rnelson0:
@Andor You didn't specify what the issue is that is being resolved, can you explain why you think this should not be present?

Andor:
@rnelson0 sorry for late response. `git` class isn't used anywhere in this module and can be removed. It is just not needed.

witchbutter:
I agree this should be removed.  The git package is likely to be defined at a higher level when using a control repository.  It would be easier to have a fact that fails the application of the module if there is no git binary, or at least a switch parameter to remove the git requirement.

siebrand:
I ran into the same issue. The issue is that the git module is no longer maintained. It installs using package[1], and not ensure_packages, and this leads to obvious and unexpected issues. Removing this dependency would be a minor change here, as well as a major improvement.

[1] https://github.com/Aethylred/puppet-git/blob/master/manifests/init.pp#L53

PR #440, opened 2018-07-05
provide proper default for $r10k_basedir

bastelfreak:
The variable basically contains the value from the
puppet_environmentpath fact. The default is
/etc/puppetlabs/code/environments. We can safely fallback to the
default value if the fact is undef. This allows the module to work
properly during the first run.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -10,7 +10,11 @@ class r10k::params
   $sources                = undef
   $puppet_master          = true
 
-  $r10k_basedir              = $facts['puppet_environmentpath']
+  if 'puppet_environment' in $facts {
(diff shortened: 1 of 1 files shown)


manifests/params.pp
```diff
@@ -10,7 +10,11 @@
   $sources                = undef
   $puppet_master          = true
 
-  $r10k_basedir              = $facts['puppet_environmentpath']
+  if $facts['puppet_environmentpath'] {
```
vchepkov: helidbe recommends to use different construct :

`if 'puppet_environment' in $facts`

bastelfreak: @hlindberg is there a technical reason for this? Is it a performance difference? The outcome at the end is the same.

PR #442, opened 2018-07-10
Replace `exec` with `Open3.capture3` to allow subsequent code to run

### LINKED ISSUE #441: Using exec in run_command replaces process and skips notify_* calls
The `run_command` function uses `exec` with a backgrounded command (using `&`) to run the desired command, which cannot get the exit status of the command, and also replaces the running webhook binary (forked from the parent) with this command.

This means that anything expected to run after calling `run_command` no longer works.

Not sure if the affected Puppet, Ruby, OS etc versions are required, but here's what I'm using currently:
 - Puppet : 4.10.12
 - Ruby : 2.1.9p490 (provided by Puppet PC1 install)
 - Distribution : CentOS 7.5
 - Module version : 6.6.0 (latest at time of writing)

How to reproduce:
 - Use slack notifications and upgrade to puppet-r10k 6.6.0

What are you seeing:
 - No Slack notifications of r10k updates

Expected behaviour:
 - Slack notifications of r10k run with success/failure status

legooolas:

This change replaces the `exec` with `Open3.capture3` as was previously used before the fork was moved to higher up in the call stack (PR #425).

Fixes #441

### DIFF
--- a/templates/webhook.bin.erb
+++ b/templates/webhook.bin.erb
@@ -267,7 +267,7 @@ class Server < Sinatra::Base
 
         message = run_command(command)
         $logger.info("message: #{message} environment: #{environment}")
-        status_message = {:status => :success, :message => message.to_s, :environment => environment, :status_code => 200}
+        status_message = {:status => :success, :message => message.to_s, :environment => environment, :status_code => 202}
         notify_slack(status_message) if slack?
       rescue => e
         $logger.error("message: #{e.message} trace: #{e.backtrace}")
@@ -318,14 +318,14 @@ class Server < Sinatra::Base
         title: "r10k deployment of Puppet environment #{target}"
       }
 
-      case status_message[:status_code]
-      when 200
+      case status_message[:status]
+      when :success
         message.merge!(
           color: "good",
           text: "Successfully deployed #{target}",
           fallback: "Successfully deployed #{target}"
         )
-      when 500
+      when :fail
         message.merge!(
           color: "bad",
           text: "Failed to deploy #{target}",
@@ -365,14 +365,14 @@ class Server < Sinatra::Base
         title: "r10k deployment of Puppet environment #{target}",
       }
 
-      case status_message[:status_code]
-      when 200
+      case status_message[:status]
+      when :success
         attachments.merge!(
           color: "good",
           text: "Successfully deployed #{target}",
           fallback: "Successfully deployed #{target}",
         )
-      when 500
+      when :fail
         attachments.merge!(
           color: "bad",
           text: "Failed to deploy #{target}",
--- a/templates/webhook.bin.erb
+++ b/templates/webhook.bin.erb
@@ -253,9 +253,9 @@ class Server < Sinatra::Base
         # the other one to complete.
         file.flock(File::LOCK_EX)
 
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hey @legooolas, thanks for the PR. Can you take a look at the used email address in the commits? It isn't associated with your github account.

legooolas:
Added the email address so that it's associated with this account now  :)

binford2k:
I am +1 to both of these changes, but would prefer they were in separate PRs

legooolas:
Left this as the `Open3`/`exec` fix and split the status code updates into PR #444

legooolas:
Er, did I break the checks with that simple change?

dhollinger:
@legooolas The change to run command was done due to the fact that if a user's Control repo is large enough, the webhook will timeout (Github and Gitlab return failures if the webhook response takes more than 10 seconds), which will cause problems for any tasks that expect a success response back from the webhook application.

When you use the `Open3.capture3`, it forces the system sending the REST call to remain connected until the command on the backend completes, which will lead to the timeout issue I mentioned above (on sufficiently large control repos). What probably needs to happen is a REST call be sent by the Webhook script (and ported into the [puppet_webhook](https://github.com/voxpupuli/puppet_webhook) gem) after a scheduled task completes.

The only reason I bring this up is that this change will break our platform that relies on the script the deploy new code automatically from a web UI. The control repo is fairly large and will case insta HTTP Failures if the `Open3.capture3` is added back in.

legooolas:
@dhollinger I'll update the patch to make `Open3.capture3` optional as a first step.
Do you think that it would make sense to have the Slack and other notifications still as part of this webhook, but with a different payload, so that whatever asyncronous tool (mcollective or r10k directly) could hit a URL once it completes and have the notification done on its behalf, or that it would make more sense for it to be part of an r10k post-run script (and hence remove notifications from this webhook entirely)?

I'm happy to update to use either option, as I'd find this useful  :)

bastelfreak:
The build errors here are caused by a broken bundler version. Please rebase against our latest master, we've fixed the issues there.

sapakt:
@dhollinger run_command method is called only in detached processes. Is it necessary to run r10k additionally in the background since main process isn't waiting for the forked ones? Change from the PR works for me - I get almost immediately 200 from gitlab, while environment is created in approximately 30 seconds later.

sapakt:
@dhollinger @bastelfreak I've checked again and run_command is executed only in `Process.detach(fork { deploy()/deploy_module })`. So it shouldn't break anything if `exec "#{command} &" ` is replaced by `stdout, stderr, exit_status = Open3.capture3(command)`. Or am I missing something? Shall I create new PR, if this one needs rebase? Those slack notifications would be really useful.

bastelfreak:
Can you checkout this PR and rebase it, or provide the fixes in a completely new PR?

sapakt:
@bastelfreak Sure. I've create new one - https://github.com/voxpupuli/puppet-r10k/pull/457.

0x6d617474:
This change is included in the now merged #485

rnelson0:
Closing as it appears to be resolved. If some parts are still needed, it will definitely need some rebasing/cleanup before re-opening.

PR #446, opened 2018-07-27
Notify rocketchat in rescue of deploy environments

amateo:
When there was an error deploying an environemnt, rocketchat was not notified (but slack was) even if rocketchat notification was configured.
This commit fixes this problem.


bastelfreak:
Thanks for the fix @amateo !

PR #452, opened 2018-09-06
Handle environment name normalized by r10k

sapakt:
This PR adds environment name normalization for <puppet generate types> command in webhook same way as r10k does.


bastelfreak:
Thanks @sapakt !

Issue #293, opened 2016-11-07
Support running `puppet generate types` with r10k mcollective plugin?


alexjfisher:
I currently use this module's mcollective plugin to run R10K on multiple puppet servers.  Is this plugin the right place to add support for running `puppet generate types` at the same time? https://docs.puppet.com/puppet/4.7/reference/environment_isolation.html#suggested-workflow

ghoneycutt:
+1 though I think it should be optional and default to false since it is both new and still experimental.

rnelson0:
Yes. Currently you could run it as the optional 'after' command. We can look at better ways to integrate it based on version, though. I just peeked at the documents, would this be for all 4.x or just starting with a specific version?

alexjfisher:
@rnelson0 The 'after' command? Or perhaps you were referring to this option? https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/configuration.mkd#postrun

The `generate types` feature is only available since puppet 4.7

rnelson0:
Yes, `postrun` (`r10k::postrun` in hiera) is the parameter whose name escaped me.

I assume we would look for `puppet_version` and use `versioncmp` on it, if it's above 4.6.0 `generate types` could be called. Sound correct?

alexjfisher:
The issue I see with `postrun` is that it's not per environment.  It's only run if you're refreshing all environments.  Maybe it's actually a patch to R10k that's needed before the R10K mcollective agent can be modified to support the additional functionality.

rnelson0:
@alexjfisher IIRC from IRC, you were investigating the patch to r10k option. Any update?

vchepkov:
+1, every time some custom type changes parameters, one has to restart puppetserver to fix an environment

saz:
`r10k` supports running `generate types`:
```
deploy:
  generate_types: true
```

Is this solving this issue?

PR #458, opened 2018-11-15
Add generate_types to Mcollective R10K plugin 

stanleyz:
This MR includes the following:

1. generate_types action to Mcollective R10K plugin
2. feature to kick off generate_types action from the webhook binary
3. Fix process fork issue introduced in previous commit 90da3cf

### DIFF
--- a/files/agent/r10k.ddl
+++ b/files/agent/r10k.ddl
@@ -127,7 +127,7 @@ action 'deploy_module', :description => "Deploy a specific module" do
   display :always
 end
 
-action 'generate_types', :description => "Generate metadata files" do
+action 'generate_types', :description => "Generate metadata files of resource types" do
   input :environment,
         :prompt => "Specific environment",
         :description => "Generate metadata files for a particular environment",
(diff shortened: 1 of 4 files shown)


stanleyz:
#293

stanleyz:
is that because the user doesn't have permission to read puppet.conf?

vchepkov:
That's a strange puppet's feature. You can check under unprivileged account:

```
$ puppet config print config
$ puppet config print|grep dir
```

They all will be in $HOME/.puppetlabs

stanleyz:
yes, so I guess you could put the additional options to the puppet.conf of unprivileged users?

Of course we can change the code to allow additional parameters, but the additional options are global options not only related to `generate types` alone.

jakegroves:
Is this still being looked into rebasing and merging?

stanleyz:
No, sorry, I've been busy on something else.

igalic:
can we transfer what's here (also) into a bolt task?

bastelfreak:
I think we can close this now because r10k supports the type generation.

PR #460, opened 2018-11-23
Updates r10k location

Rudikza:

The path for r10k.yaml inside /usr/local/bin/prefix_command.rb is outdated.  This commit changes the path. I did think about making it a param but I see in  https://github.com/voxpupuli/puppet-r10k/issues/342 that the path was just hardcoded so I did the same.

### DIFF
--- a/files/prefix_command.rb
+++ b/files/prefix_command.rb
@@ -5,7 +5,7 @@ require 'yaml'
 if STDIN.tty?
   puts 'This command is meant be launched by webhook'
 else
-  prefixes = YAML.load_file('/etc/r10k.yaml')
+  prefixes = YAML.load_file('/etc/puppetlabs/r10k/r10k.yaml')
 
   json_data = JSON.parse(STDIN.read)
   url = json_data['repository']['url']


alexjfisher:
Since there's already a parameter for `configfile`, I think it'd be a shame not to use it.
If you don't, this change *could* be breaking for some users.  eg.
```puppet
class { 'r10k':
  configfile                => '/some/nonstandard/path/to/r10k.yaml',
  manage_configfile_symlink => true, # Will link /etc/r10k.yaml to path above
}
```

Rudikza:
@alexjfisher I see your point but in the code, $configfile points to $r10k_config_file which is set to `/etc/puppetlabs/r10k/r10k.yaml`. 

The only places in the whole codebase that still use the old  path `/etc/r10k.yaml` are in the docs, as a value for configfile_symlink, in the tests and hardcoded in the `prefix_command.rb` file.  

Perhaps a good compromise would be to change the prefix_command.rb file to a template and then use $configfile to set the path?

bastelfreak:
@Rudikza thanks for your contribution. Always happy to see people with gpg signed commits \o/

alexjfisher:
> Perhaps a good compromise would be to change the prefix_command.rb file to a template and then use $configfile to set the path?

That's exactly what I had in mind. :)

alexjfisher:
My bad.  I didn't realised that this file was provided as an 'example' prefix command file.  As such, I think this simple change is the only sensible one.

PR #461, opened 2018-11-23
Remove `install_gcc` parameter

Rudikza:
Currently r10k requires puppetlabs/gcc which hasn't been updated in almost 4 years in turns only allows up to stdlib version <5.0.0. There are certain features which we would really like to use in 5x but can't. 
I'm not too sure about upgrading GCC but while I was poking around in the code base I found the following comment which indicated that GCC should have been removed once r10k it version 1.0
```
 # There are currently bugs in r10k 1.x which make using 0.x desireable in
  # certain circumstances. However, 0.x requires make and gcc. Conditionally
  # include those classes if necessary due to 0.x r10k version usage. When
  # 1.x is just as good or better than 0.x, we can stop supporting 0.x and
  # remove this block.
```
I'm hoping that since r10k is now at version 3 and no longer requires GCC or make that this PR will be accepted.  Alternatively we could try and get GCC moved to voxpupuli and I could have a go at upgrading but this seems like a much shorted path to victory :smile:

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -12,15 +12,20 @@ By participating in this project you agree to abide by its terms.
 
 1. Create a separate branch for your change.
 
-1. Run the tests. We only take pull requests with passing tests, and
-   documentation.
+1. We only take pull requests with passing tests, and documentation. [travis-ci](http://travis-ci.org)
+   runs the tests for us. You can also execute them locally. This is explained
+   in a later section.
+
+1. Checkout [our docs](https://voxpupuli.org/docs/#reviewing-a-module-pr) we
+   use to review a module and the [official styleguide](https://puppet.com/docs/puppet/6.0/style_guide.html).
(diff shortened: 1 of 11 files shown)


vchepkov:
make should be removed as well in this case, IMHO

.github/CONTRIBUTING.md
```diff
@@ -12,15 +12,20 @@ By participating in this project you agree to abide by its terms.
 
 1. Create a separate branch for your change.
 
-1. Run the tests. We only take pull requests with passing tests, and
-   documentation.
+1. We only take pull requests with passing tests, and documentation. [travis-ci](http://travis-ci.org)
```
bastelfreak: Why is this file changed in the PR? Did you also do a modulesync?

bastelfreak:
Hi @Rudikza, thanks for the PR. Can you please take a look at the used email address in the commit? It isn't associated with your github account.

PR #465, opened 2018-11-28
Removes gcc and make

Rudikza:
Currently, r10k requires puppetlabs/gcc which hasn't been updated in almost 4 years in turns only allows up to stdlib version <5.0.0. There are certain features which we would really like to use in 5x but can't.
I'm not too sure about upgrading GCC but while I was poking around in the code base I found the following comment which indicated that GCC should have been removed once r10k it version 1.0

```
 # There are currently bugs in r10k 1.x which make using 0.x desireable in
  # certain circumstances. However, 0.x requires make and gcc. Conditionally
  # include those classes if necessary due to 0.x r10k version usage. When
  # 1.x is just as good or better than 0.x, we can stop supporting 0.x and
  # remove this block.
```
This is a breaking change but according to the puppet-r10k version chart, the current version of puppet-r10k supports the r10k v.3.0.3 gem which no longer requires gcc or make so I dont think this should be a blocker.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -2,9 +2,7 @@ fixtures:
   repositories:
     stdlib:  "https://github.com/puppetlabs/puppetlabs-stdlib.git"
     ruby:    "https://github.com/puppetlabs/puppetlabs-ruby.git"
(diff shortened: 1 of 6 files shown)


bastelfreak:
Thanks!

PR #466, opened 2018-11-29
WIP: This PR removes the hard-coded value for the prefix_command configfile

Rudikza:
file value.

This PR removes the hard-coded value for the prefix_command  by moving the prefix_command file resource to a template and then using the $configfile param which is defaulted to `/etc/puppetlabs/r10k/r10k.yaml`.

### DIFF
--- a/examples/webhook/config.pp
+++ b/examples/webhook/config.pp
@@ -1,11 +1,11 @@
 include ::r10k::webhook
 
 file {'/usr/local/bin/prefix_command.rb':
-  ensure => file,
-  mode   => '0755',
-  owner  => 'root',
-  group  => '0',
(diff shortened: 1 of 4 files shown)


Rudikza:
Still busy testing . . . .

Rudikza:
@alexjfisher This is going to be more complicated that I thought.  At the moment the prefix_command file is called from outside the module. ` source => 'puppet:///modules/r10k/prefix_command.rb'` when I moved this to a template it can no longer be called outside the module because the $configfile param is not available.

I could add in a new param called manage_prefix_command_file and then have that call the template. That might be a better solution because we can keep the old method and introduce the new method. Maybe enable it by default?

PR #469, opened 2018-12-01
enable acceptance tests

Dan33l:

First work to enable beaker acceptance tests in the module.

Code to make tests was already present. It expected some output. And now i get no output at all.

### DIFF
--- a/spec/acceptance/basic_webhook_spec.rb
+++ b/spec/acceptance/basic_webhook_spec.rb
@@ -1,9 +1,8 @@
 require 'spec_helper_acceptance'
-require 'json'
+
 describe 'System Ruby with No SSL, Not protected, No mcollective' do
   context 'default parameters' do
-    let(:pp) do
-      "
+    pp = %(
       class { 'r10k':
         remote => 'git@github.com:someuser/puppet.git',
       }
@@ -17,8 +16,7 @@ describe 'System Ruby with No SSL, Not protected, No mcollective' do
       class {'r10k::webhook':
         require => Class['r10k::webhook::config'],
       }
-      "
-    end
+    )
 
     it 'applies with no errors' do
       apply_manifest(pp, catch_failures: true)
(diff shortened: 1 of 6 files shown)


bastelfreak:
@Dan33l should we merge this as is or do you want to work more on it?

Dan33l:
We could merge as this. 
But i am interested by understanding why the original code expect some output. Now, when i check manually same commands i get no output at all.
If it's because the tested software has evolved, we should probably have the tests evolve as well.

spec/acceptance/basic_webhook_spec.rb
```diff
@@ -1,9 +1,8 @@
 require 'spec_helper_acceptance'
-require 'json'
+
 describe 'System Ruby with No SSL, Not protected, No mcollective' do
   context 'default parameters' do
-    let(:pp) do
-      "
+    pp = %(
```
ekohl: Why should this not be a let block anymore?

spec/spec_helper_acceptance.rb
```diff
@@ -1,33 +1,22 @@
...
-      JSON.parse(File.read('metadata.json'))['dependencies'].each do |dependency|
-        on host, puppet('module', 'install', dependency['name'].to_s), acceptable_exit_codes: [0, 1]
-      end
+    hosts.each do |_host|
+      ensure_puppet_user_exists = %(
+        user { 'puppet':
+          ensure => present,
+        }
+      )
+      apply_manifest(ensure_puppet_user_exists, catch_failures: true)
```
ekohl: A shorter way to write this is `on host, puppet('resource', 'user', 'ensure=present')`. Otherwise you should `apply_manifest_on(host, ensure_puppet_user_exists, catch_failures: true)` for it to work correctly on multiple nodes

Dan33l:
Since the keyword `success` is not in the output, i added a check of missing `You shall not pass`.

spec/acceptance/basic_webhook_spec.rb
Dan33l: IMO, it is clearer in this way.

spec/spec_helper_acceptance.rb
Dan33l: With this shorter way i got this log : 
```
centos7-64-1 23:05:44$ puppet resource user ensure=present
  user { 'ensure=present':
    ensure => 'absent',
  }
```
And even with `puppet resource user puppet`, it is yet produced `ensure=>absent`. hu!

spec/acceptance/basic_webhook_spec.rb
```diff
@@ -30,38 +28,32 @@ class {'r10k::webhook':
...
-      it { expect(r.stdout).to match(%r{^.*success.*$}) }
-      it { expect(r.exit_code).to eq(0) }
+    describe command('/usr/bin/curl -X POST -d \'{ "push": { "changes": [ { "new": { "name": "production" } } ] } }\' "http://localhost:8088/payload" -q') do
+      its(:stdout) { is_expected.not_to match %r{.*You shall not pass.*} }
+      its(:exit_status) { is_expected.to eq 0 }
     end
-    # rubocop:enable RSpec/RepeatedExample
 
-    # rubocop:disable RSpec/MultipleExpectations
-    it 'successfully locks when hammered with multiple requests' do
```
ekohl: I do wonder about the concurrency here. describe blocks are run separate so you might be changing the test to run 4 commands sequentially while the test is actually run 4 commands in parallel.

spec/acceptance/basic_webhook_spec.rb
ekohl: There is a difference in evaluation and https://stackoverflow.com/questions/5359558/when-to-use-rspec-let/5359979#5359979 outlines why let blocks are preferred.

spec/spec_helper_acceptance.rb
ekohl: Oh, it should have been `on host, puppet('resource', 'user', 'puppet', 'ensure=present')`

ekohl: And you should have at least looked at my other comment. This current version is applying the same manifest only on the default node for every host. In a single node env that doesn't matter, but it's a bad example and people copy paste.

PR #475, opened 2018-12-08
set webrick and sinatra versions to 'installed'

### LINKED ISSUE #471: don't pin sinatra and webrick gems anymore

- Puppet: 6.0
- Distribution: Debian 9.6
- Module version: master

## What are you seeing

Now that WEBrick is in the Ruby standard library, and Puppet Platform 6.0 packages include Ruby 2.5.1, we shouldn't need to pin the webrick and sinatra versions anymore. So I believe these commits could just be reverted:
* 312e759dbfae9b60270a6e03827762f15a2cb9db
* d36db9a249542709a9149ea4af40be4f6596549f

kenyon:
Since webrick has been part of the ruby standard library for a while, its version doesn't need to be specified anymore.

The sinatra gem comes with Puppet, so its version also doesn't need to be specified.

Fixes  #471.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -157,8 +157,12 @@ class r10k::params
   $webhook_configfile_mode       = '0644'
   $webhook_ignore_environments   = []
   $webhook_mco_arguments         = undef
-  $webhook_sinatra_version       = '~> 1.0'      # Sinatra 2 requires rack 2 which in turn requires ruby 2.2. Puppet 4 AIO ships with ruby 2.1
-  $webhook_webrick_version       = '1.3.1'       # Webrick 1.4 requires ruby >= 2.3
+  if $facts['pe_server_version'] == '2016.4.2' {
+    $webhook_sinatra_version       = '~> 1.0'
+  } else {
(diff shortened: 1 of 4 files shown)


manifests/params.pp
```diff
@@ -157,8 +157,12 @@
   $webhook_configfile_mode       = '0644'
   $webhook_ignore_environments   = []
   $webhook_mco_arguments         = undef
-  $webhook_sinatra_version       = '~> 1.0'      # Sinatra 2 requires rack 2 which in turn requires ruby 2.2. Puppet 4 AIO ships with ruby 2.1
-  $webhook_webrick_version       = '1.3.1'       # Webrick 1.4 requires ruby >= 2.3
+  if $facts['pe_server_version'] == '2016.4.2' {
```
bastelfreak: Do we still support this version, or is it already EoL? Also you could use a selector statement for this. https://puppet.com/docs/puppet/5.3/lang_conditional.html#selectors

bastelfreak: okay, lets keep it for now.

PR #479, opened 2019-01-31
Use latest gentoo/portage release

ghoneycutt:
This release was previously pinned because of bugs in 2.4.0. The next
release, 2.4.5, claims to have fixed these issues. Setting the lower
bound to 2.4.5 as to not include the known bad version of 2.4.0.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -6,6 +6,4 @@ fixtures:
     inifile: "https://github.com/puppetlabs/puppetlabs-inifile.git"
     vcsrepo: "https://github.com/puppetlabs/puppetlabs-vcsrepo.git"
     git:     "https://github.com/puppetlabs/puppetlabs-git.git"
-    portage:
-      repo: "https://github.com/gentoo/puppet-portage.git"
-      ref: "2.3.0"
+    portage: "https://github.com/gentoo/puppet-portage.git"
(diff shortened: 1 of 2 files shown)


bastelfreak:
I'm a bit confused here. Didn't they release their latest versions on the forge?
https://github.com/gentoo/puppet-portage/releases lists2.4.5 , but https://forge.puppet.com/gentoo/portage lists only 2.3.0 :(

ghoneycutt:
Seems they have not released it to the forge, though they have a github release. @bastelfreak glad you spoke up, since I saw you did the original pinning. :)

ghoneycutt:
gentoo/portage looks abandoned and Gentoo is not even listed in metadata. Perhaps it is time to just pull all the Gentoo code out.

bastelfreak:
since the code doesn't cause any harm, should we only remove it from the metadata.json? Makes it easier for people in the future to properly add Gentoo support.

ghoneycutt:
Gentoo is already not listed in the metadata, so no support for that platform is given. Leaving code around creates more work in the future, so would we should just remove the code.

bastelfreak:
ah yeah, :+1:

ghoneycutt:
@bastelfreak could you review https://github.com/voxpupuli/puppet-r10k/pull/480

PR #482, opened 2019-04-07
modulesync 2.7.0 and drop puppet 4

bastelfreak:
modulesync 2.7.0


spec/spec_helper.rb
```diff
@@ -1,10 +1,23 @@
 # This file is managed via modulesync
 # https://github.com/voxpupuli/modulesync
 # https://github.com/voxpupuli/modulesync_config
+RSpec.configure do |c|
+  c.mock_with :rspec
```
rnelson0: Is this why the stub is failing, was this expected to be `:mocha`? I am not well versed in mocks but seem to recall that being the default previously.

dhoppe: I did a rewrite of the test #484 based on https://github.com/voxpupuli/puppet-autofs/pull/146/commits/491d6fc5a53d56897fc7bf50f532a2cccbf21f74. Unfortunately I do not have write permissions to this repository, so I was forced to create an additional pull request.

PR #485, opened 2019-05-06
Webhook improvements

### LINKED ISSUE #441: Using exec in run_command replaces process and skips notify_* calls

### LINKED ISSUE #472: webhook swallowing r10k error messages

- Puppet: 5.5.6
- Ruby: 2.4.4p296
- Distribution: RHEL 7.5
- Module version: 6.7.1-rc0

## What are you seeing

I had an issue where the webhook was not fetching/building the puppet forge modules. There was nothing in the webhook access.log and when I ran r10k from the command line it worked perfectly.  After much digging (and using pry-remote) I found that r10k was being blocked by our proxy server.  

The proxy is a quick fix but I was more concerned that the webhook app seemed to swallow the errors from r10k. 

I found that this block of code seemed to the cause of the problem:
```ruby
    def run_command(command)
      message = ''
      File.open(LOCKFILE, 'w+') do |file|
        # r10k has a small race condition which can cause failed deploys if two happen
        # more or less simultaneously. To mitigate, we just lock on a file and wait for
        # the other one to complete.
        file.flock(File::LOCK_EX)

        message = "triggered: #{command}"
        exec "#{command} &"
        exit_status = 0
        raise "#{stdout}\n#{stderr}" if exit_status != 0
      end
      message
    end
```
I played around a bit with  Open3.capture2e and came up with the following code which correctly picks up the failure, logs the error messages to the log file and then raises a failure.
```ruby
    def run_command(command)
      message = ''
      File.open(LOCKFILE, 'w+') do |file|
        # r10k has a small race condition which can cause failed deploys if two happen
        # more or less simultaneously. To mitigate, we just lock on a file and wait for
        # the other one to complete.
        file.flock(File::LOCK_EX)
        message = "triggered: #{command}"
        stderr, status = Open3.capture2e(command)
        unless status.success?
          ar = stderr.split("\n").grep /ERROR/
          ar.each { |e| $logger.error(e)}
          "#{command} failed to execute"
          raise "Failed to exec  #{command}"
        end
      end
      message
    end
```
The logfile formatting needs a bit of work  (https://gist.github.com/Rudikza/5d519b206f9879f25796f5b3e71dce6e) and we should perhaps look at adding a error log file? I'm also not sure if the raise is the correct way to handle this type of error.

Would this be worth adding to the module and spending more time on?

0x6d617474:
This PR improves the webhook functionality in the following areas: 
 - Correct HTTP status codes returned for the `/payload` and `/module` endpoints
 - Allow chat notifications (Slack/Rocketchat) to execute correctly
 - Correct ordering of chat notifications (Slack/Rocketchat)
 - Update chat messages for better accuracy on the operation
 - Add support for customizing the message icon in Slack

Each commit message contains detailed reasons for the changes.

Fixes #472 
Fixes #441


0x6d617474:
Note, I don't have access to a Rocketchat instance to test against currently, but I have no reason to suspect anything should have broken. 

I have successfully tested the Slack functionality, and it works as expected. Here is an example using a customized icon: 

image

binford2k:
@dhollinger do we want to transition to voxpupuli/puppet_webhook at some point?

PR #487, opened 2019-05-09
Add support for proxy settings

dhoppe:
This closes #473. @Rudikza Thank you very much for this enhancement.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -21,6 +21,9 @@
 #   Location of symlink that points to configfile. Default: /etc/r10k.yaml
 # * [*forge_settings*]
 #   Hash containing settings for downloading modules from the Puppet Forge.
+# * [*proxy*]
+#   String containing proxy setting for r10k.yaml.
+#   Default: ''
 #
 # === Examples
 #
@@ -62,6 +65,7 @@ class r10k::config (
   $root_user                                = $r10k::params::root_user,
   $root_group                               = $r10k::params::root_group,
   Stdlib::Absolutepath $puppetconf_path     = $r10k::params::puppetconf_path,
+  Optional[String[1]] $proxy                = $r10k::params::proxy,
   String $r10k_yaml_template                = 'r10k/r10k.yaml.erb',
 ) inherits r10k::params {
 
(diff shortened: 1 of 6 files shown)


manifests/params.pp
```diff
@@ -9,6 +9,7 @@
   $install_options        = []
   $sources                = undef
   $puppet_master          = true
+  $proxy                  = ''
```
bastelfreak: please default to undef and not empty string

manifests/config.pp
```diff
@@ -62,6 +65,7 @@
   $root_user                                = $r10k::params::root_user,
   $root_group                               = $r10k::params::root_group,
   Stdlib::Absolutepath $puppetconf_path     = $r10k::params::puppetconf_path,
+  String $proxy                             = $r10k::params::proxy,
```
bastelfreak:
```suggestion
  Optional[String[1]] $proxy                = $r10k::params::proxy,
```

manifests/config.pp
```diff
@@ -21,6 +21,9 @@
 #   Location of symlink that points to configfile. Default: /etc/r10k.yaml
 # * [*forge_settings*]
 #   Hash containing settings for downloading modules from the Puppet Forge.
+# * [*proxy*]
+#   String containing proxy setting for r10k.yaml.
+#   Default: ''
```
Dan33l: `params.pp` says it defaults to `undef`.

dhoppe: It does not matter. I will create another pull request, which will refactor a lot of stuff and hopefully we will use data-in-modules and generate a REFERENCE.md.

manifests/init.pp
```diff
@@ -21,6 +21,7 @@
   Hash $forge_settings                                        = $r10k::params::forge_settings,
   Hash $deploy_settings                                       = $r10k::params::deploy_settings,
   $root_user                                                  = $r10k::params::root_user,
+  Optional[String[1]] $proxy                                  = $r10k::params::proxy,
```
alexjfisher: `Stdlib::HTTPUrl` instead of `String[1]`?

PR #488, opened 2019-05-10
[WIP] Refactor classes postrun_command, prerun_command

dhoppe:
This pull request will prepare the module for data-in-modules and provide a REFERENCE.md. Unfortunately this module uses a lot of classes and I need to figure out which one are private. Because some variables are named or used badly, this change will be backwards incompatible and end up in a major release.


data/common.yaml
```diff
@@ -0,0 +1,8 @@
+---
+r10k::include_postrun_command: false
+r10k::postrun_command::ensure: 'present'
+r10k::postrun_command::command: 'deploy environment -p'
```
jcpunk, line 4: `r10k` is showing `-p` is deprecated, and suggests use of `-m`

Issue #449, opened 2018-08-23
yum provider refers to ghost package


ggeldenhuis:
If you set the provider attribute to yum it tries and install a package called **rubygem-r10k**. However that package does not exist nor can I find any reference for creating the package.

bastelfreak:
Hi @ggeldenhuis. There are no rpms as far as I know. At least the Arch Linux people are working on packages, but there are no available for RHEL platform. You need to create your own or stick with the gem/puppet_gem provider.

dhoppe:
@ggeldenhuis This piece of code is almost 6 years old and was introduced with version 1.0.0. At this time RHEL 5 and SLES 11.3 provided a ```rubygem-r10k``` package. 

I will create a pull request to fix this bug.

ggeldenhuis:
Thanks!

jorhett:
This appears to be fixed?

PR #489, opened 2019-05-10
Remove support for providers yum, zypper

dhoppe:
This pull request removes support for providers ```yum``` and ```zypper```. We also might want to remove support for OpenBSD and remove the ```package``` resource, since this was introduced for Puppet Enterprise which has ```r10k``` already installed.

This will fix issue #449.

### DIFF
--- a/README.md
+++ b/README.md
@@ -781,7 +781,6 @@ The supported installation modes for this module
 * bundle
 * puppet_gem
 * gem
-* zypper
 
 ##### `install_options`
 Options to pass to the `provider` declaration
--- a/README.md
+++ b/README.md
@@ -778,7 +778,6 @@ The name of the package to be installed via the provider
 ##### `provider`
 The supported installation modes for this module
 
-* yum
 * bundle
 * puppet_gem
 * gem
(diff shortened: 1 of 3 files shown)


alexjfisher:
This package https://software.opensuse.org/download.html?project=systemsmanagement%3Apuppet&package=rubygem-r10k isn't that old...

dhoppe:
@alexjfisher Hm. I missed that somehow and will drop the latest commit. Interesting that SLES provides a package and RHEL does not. Usually it is the other way around.

rnelson0:
r10k 2.3.0 was released on May 18, 2016 (current is 3.2.0 from April 16, 2019), so I'm not sure that I would say the packages are anything near recent, even if the timestamps are.

dhoppe:
Someone needs to make a decision. I still think we should remove yum and zypper, because it does not hurt to install a gem.

bastelfreak:
let's keep it open for 3 more days. I'm :+1: for dropping it. If nobody offers an argument to keep it, we can drop it.

PR #494, opened 2019-06-14
webhook: use the systemd service by default

kenyon:
systemd is standard nowadays, so default to that instead of the init script.

Remove the suse service template because the redhat one would work equivalently. Rename the redhat one to be generic.


rnelson0:
While systemd is standard, we still support EL6 and others where systemd is not in use. You might be able to share the systemd service file between EL7 and SuSE but you will definitely need to make sure sysvinit systems other than gentoo continue to work.

kenyon:
Yeah, I did forget about EL 6. Updated to support that.

PR #495, opened 2019-06-17
Wrap slack_icon with colons

0x6d617474:
The configuration file for the webhook config writes values that
begin with colons (:) as symbols rather than strings, to allow
for calculated values.

Since slack emoji names follow the pattern of `:name:`, they are
written as symbols instead of strings, and the resulting config
file breaks the webhook service.

This change simply requires slack icons to be entered without the
wrapping colons, and adds them back just before sending to Slack.

Where previously the ocean icon would be written as `:ocean:` in
the webhook_slack_icon parameter, now it would be written as `ocean`.

n/a

### DIFF
--- a/templates/webhook.bin.erb
+++ b/templates/webhook.bin.erb
@@ -300,7 +300,7 @@ class Server < Sinatra::Base
       end
 
       if $config['slack_icon']
-        slack_icon = $config['slack_icon']
+        slack_icon = ":#{$config['slack_icon']}:"
       else
         slack_icon = ':ocean:'
       end


rnelson0:
Does this deserve a little disclaimer somewhere in the docs, so people know to adjust?

0x6d617474:
I don't think the `slack_icon` feature is mentioned anywhere in the documentation, and if you tried to use it before this patch you'd get a broken service, so I don't think this change will affect anyone negatively.  

I can use this PR to update the docs to include the `slack_icon` feature if you'd like.

rnelson0:
Since it was not documented already, I went ahead and merged it prior to 7.0.0, but please do open a documentation PR when you get a chance!

PR #502, opened 2019-10-23
remove git dependency from r10k::install::puppet_gem

Andor:

Copy of https://github.com/voxpupuli/puppet-r10k/pull/435


bastelfreak:
thanks for the PR!

alexjfisher:
I agree that this didn't belong in this module.  However, I think this should have been marked as breaking and documented in the README.

alexjfisher:
The better option might have been to introduce a `manage_git` parameter.

PR #505, opened 2019-11-14
allow puppetlabs/inifile 4.x

bastelfreak:


vchepkov:
@bastelfreak , @alexjfisher, I think we addressed failing tests in #513, this should succeed now.
Any chance to get a release soon?

PR #511, opened 2020-03-28
Fixes: Travis CI tests, webhook IPv4/IPv6, webhook intermediate certificates

mcb30:
#### Pull Request (PR) description
This pull request fixes three issues:
- the currently failing Travis CI tests (#509)
- support for the webhook to listen on both IPv4 and IPv6 by default (#490)
- support for the webhook to use intermediate TLS certificates (#510)

These three issues are unrelated.  They are submitted as a single pull request only because the first commit in this series is required to fix the Travis CI tests, without which it is impossible to get a Travis CI pass for the other commits in the series.

#### This Pull Request (PR) fixes the following issues
Fixes #509 
Fixes #490 
Fixes #510

PR #512, opened 2020-03-28
fix webhook generate type functionailty for non-root user

### LINKED ISSUE #412: generate types feature doesn't work under unprivileged user id

vchepkov:
if webhook is running under unprivileged account it creates
.resource_types in it's home directory, instead of code directory
explicitly pass parameters to point to configured code location

Fixes #412

### DIFF
--- a/manifests/webhook.pp
+++ b/manifests/webhook.pp
@@ -49,7 +49,7 @@ class r10k::webhook(
     ensure => $ensure_file,
     owner  => $user,
     group  => $group,
-    before => File['webhook_bin'],
+    before => Service['webhook'],
   }
 
   file { '/var/log/webhook':
@@ -57,14 +57,14 @@ class r10k::webhook(
     owner  => $user,
     group  => $group,
     force  => $ensure,
-    before => File['webhook_bin'],
+    before => Service['webhook'],
   }
 
   file { '/var/run/webhook':
     ensure => $ensure_directory,
     owner  => $user,
     group  => $group,
-    before => File['webhook_init_script'],
+    before => Service['webhook'],
   }
 
   file { 'webhook_init_script':
@@ -72,7 +72,7 @@ class r10k::webhook(
     content => template("r10k/${service_template}"),
     path    => $service_file,
     mode    => $service_file_mode,
-    before  => File['webhook_bin'],
+    notify  => Service['webhook'],
   }
 
   file { 'webhook_bin':
--- a/templates/webhook.bin.erb
+++ b/templates/webhook.bin.erb
@@ -273,7 +273,16 @@ class Server < Sinatra::Base
 
     def generate_types(environment)
       begin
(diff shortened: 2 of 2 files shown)


vchepkov:
I think tests are failing due to this recent change in systemd https://access.redhat.com/solutions/4420581

I am not sure how to address it yet, open for suggestions

vchepkov:
K, I fixed the systemd unit file, so tests don't fail on systems with new systemd
@bastelfreak , @alexjfisher , could you review, please?

bastelfreak:
looks good to me, but it would be nice if somebody else can review / test this locally.

templates/webhook.service.erb
```diff
@@ -3,19 +3,16 @@ Description=R10K Webhook Service
 After=syslog.target network.target
 
 [Service]
-Type=forking
+Type=simple
 EnvironmentFile=-/etc/sysconfig/webhook
 RuntimeDirectory=webhook
 User=<%= @user %>
-PIDFile=/var/run/webhook/webhook.pid
```
alexjfisher: I think this line was only a problem in CI in docker.  On real systems it would be fine.

templates/webhook.service.erb
```diff
@@ -3,19 +3,16 @@ Description=R10K Webhook Service
 After=syslog.target network.target
 
 [Service]
-Type=forking
+Type=simple
```
alexjfisher: To change this, wouldn't you also have to change the process such that it doesn't daemonise and create the pidfile?

alexjfisher: Oh, ignore that, you have.  Still not convinced we should mess around too much just because docker has an issue.

vchepkov: Actually, not just docker, it affects regular installations as well
That RedHat article I linked above recommending not using Type=forking as a legacy method.

vchepkov:
Same does systemd project itself - https://www.freedesktop.org/software/systemd/man/systemd.service.html

> It is generally recommended to use Type=simple for long-running services whenever possible, as it is the simplest and fastest option.

vchepkov:
#511 is proposing the same fix, btw

alexjfisher:
I've pulled out just the systemd change into its own PR. https://github.com/voxpupuli/puppet-r10k/pull/513

vchepkov:
Rebased

PR #515, opened 2020-03-29
Use full certificate chain for webhook

### LINKED ISSUE #510: webhook does not support the use of intermediate certificates
All versions

## How to reproduce (e.g Puppet code you use)
Attempt to use any webhook server certificate that requires the use of an intermediate certificate.  For example, using a certificate for the machine's FQDN issued by LetsEncrypt:
```
class { 'r10k::webhook::config':
  enable_ssl => true,
  public_key_path => "/etc/letsencrypt/live/${::fqdn}/fullchain.pem",
  private_key_path => "/etc/letsencrypt/live/${::fqdn}/privkey.pem",
}
class { 'r10k::webhook':
  require => Class['r10k::webhook::config'],
}
```

## What are you seeing
Connections to the webhook fail TLS verification because the intermediate certificate is missing.  For example:
```
$ gnutls-cli <fqdn> -p 8088
...
- Got a certificate list of 1 certificates.
- Status: The certificate is NOT trusted. The certificate issuer is unknown. 
```

## What behaviour did you expect instead
Connections to the webhook should succeed.  This requires the webhook to present the intermediate certificate(s) along with its own certificate.

## Any additional information you'd like to impart
A pull request to fix this issue is ready and will be submitted shortly.

mcb30:

Split from #511 

The webhook service is currently limited to using a certificate
directly issued by a trusted CA; it will silently ignore any
intermediate certificates that are present in the certificate file.

The currently released versions of the Ruby OpenSSL libraries do not
provide any clean way to load a certificate chain from a file.  We
therefore split the file using the BEGIN/END markers as per RFC 7468,
and construct the certificate chain directly.

No tests are extended to cover this enhancement, since there is no
existing test coverage for the use of HTTPS certificates by the
webhook.  All current tests use plain HTTP via http://localhost:8088.

Fixes #510

### DIFF
--- a/templates/webhook.bin.erb
+++ b/templates/webhook.bin.erb
@@ -57,10 +57,15 @@ opts = {
   :StartCallback  => Proc.new { File.open(PIDFILE, 'w') {|f| f.write Process.pid } },
 }
 if $config['enable_ssl'] then
+  CERT_PATTERN = /-----BEGIN.+?CERTIFICATE-----.+?-----END.+?CERTIFICATE-----/m
+  certs = File.open("#{$config['public_key_path']}").read.scan(CERT_PATTERN)
   opts[:SSLVerifyClient] = OpenSSL::SSL::VERIFY_NONE
-  opts[:SSLCertificate]  = OpenSSL::X509::Certificate.new(File.open("#{$config['public_key_path']}").read)
+  opts[:SSLCertificate]  = OpenSSL::X509::Certificate.new(certs.shift)
   opts[:SSLPrivateKey]   = OpenSSL::PKey::RSA.new(File.open("#{$config['private_key_path']}").read)
   opts[:SSLCertName]     = [ [ "CN",WEBrick::Utils::getservername ] ]
+  if ! certs.empty? then
+    opts[:SSLExtraChainCert] = certs.map { |x| OpenSSL::X509::Certificate.new(x) }
+  end
 end
 
 if $config['use_mcollective'] then


vchepkov:
Wouldn't it be better just provide intermediate ca in a separate file?

mcb30:
> Wouldn't it be better just provide intermediate ca in a separate file?

There may be multiple intermediate certificates, so the PEM splitting logic would still be required.

I chose to use the full-chain approach because it minimises the changes to the existing codebase and is backwards-compatible with existing configurations: if an existing configuration happens to switch to a cert file that now includes an intermediate cert then it will work with no configuration changes required.

@vchepkov I can rework the patch to require the intermediate certs to be held in a separate file (and hence separate configuration parameters, etc) if you prefer, but it will necessarily be a much more invasive patch.

vchepkov:
I was just curios and yes, multiple intermediate certificates would have the same issue.
Since webhook interface is a very specialized service and configured/called in point-to-point manner only by git hooks, this fix is only necessary if there is no option to turn off certificate verification. I imagine code will fail if certificate and key are kept in the same file (as some folks do). Why don't you split using `-----BEGIN CERTIFICATE-----.+-----END CERTIFICATE-----` ?

mcb30:
> Since webhook interface is a very specialized service and configured/called in point-to-point manner only by git hooks, this fix is only necessary if there is no option to turn off certificate verification.

In our use case, the webhook will be called by a git management service in a different administrative domain (e.g. GitHub), so there is no way to disable certificate verification.

> I imagine code will fail if certificate and key are kept in the same file (as some folks do). Why don't you split using `-----BEGIN CERTIFICATE-----.+-----END CERTIFICATE-----` ?

Can do.  RFC 7468 mentions some existing non-standard implementations that use "BEGIN X.509 CERTIFICATE".  How about a regex of `/-----BEGIN.*?CERTIFICATE-----.*?-----END.*?CERTIFICATE-----/m`?

vchepkov:
I think it should be `.+?`, but yes, otherwise you can end up with chain having only the key and it would fail

mcb30:
> I think it should be `.+?`, but yes, otherwise you can end up with chain having only the key and it would fail

Thanks.  Will modify the code and update the PR after checking that it still works in my test deployment.

mcb30:
@alexjfisher Are there any changes you'd like made, or can this be merged?  Thanks.

alexjfisher:
@mcb30 Thanks!

PR #517, opened 2020-04-02
Add support for package source attribute

prolixalias:
- source param on package type
- optional on install subclass
Allow source param on package

n/a

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -8,6 +8,7 @@ class r10k::install (
   $manage_ruby_dependency,
   $puppet_master = true,
   $is_pe_server = $r10k::params::is_pe_server,
+  Optional[String[1]] $gem_source = undef,
 ) inherits r10k::params {
 
   if $package_name == '' {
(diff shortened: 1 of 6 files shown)


dhoppe:
@prolixalias Could you please add an RSpec test that uses the new parameter?

manifests/install.pp
```diff
@@ -8,6 +8,7 @@
   $manage_ruby_dependency,
   $puppet_master = true,
   $is_pe_server = $r10k::params::is_pe_server,
+  Optional[String[1]] $gem_source = undef,
```
bastelfreak, line 11: can you please document the new parameter?

prolixalias: Sure thing, thanks. All set...

prolixalias:
@bastelfreak & @dhoppe - any interest in converting to PDK? Once this PR is through, I'd be happy to convert/adjust/PR... Otherwise, I'll go back about my regularly scheduled programming.

bastelfreak:
@prolixalias thanks for the fixes. At the moment we use modulesync (https://github.com/voxpupuli/modulesync and https://github.com/voxpupuli/modulesync_config) to manage our modules. We haven't decided yet if we migrate to pdk or not. If we do, we would need to migrate all of our modules.

PR #519, opened 2020-04-03
Remove `anchor` resource from `r10k::install::gem`

alexjfisher:
I've refactored the class to not use the anchor resource.
Since
https://github.com/voxpupuli/puppet-r10k/commit/78283bd8c1995f54398ec30478fb0adfc6ea59ec#diff-41c34a669409f2dbaa4211d5f642a003L40
it is not being referred to from any other class.

If someone needs to create a similar relationship, they can use
`Class['r10k::install::gem']` instead.

### DIFF
--- a/manifests/install/gem.pp
+++ b/manifests/install/gem.pp
@@ -4,23 +4,19 @@ class r10k::install::gem (
   $version,
 ) {
   require git
-  anchor{'r10k::ruby_done':}
+
   case $manage_ruby_dependency {
     'include': {
       include ruby
-      include ruby::dev
-      Class['::ruby']
-      -> Class['ruby::dev']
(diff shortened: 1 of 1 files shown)


manifests/install/gem.pp
```diff
@@ -4,23 +4,19 @@
...
 ) {
   require git
-  anchor{'r10k::ruby_done':}
+
   case $manage_ruby_dependency {
     'include': {
       include ruby
-      include ruby::dev
-      Class['::ruby']
-      -> Class['ruby::dev']
```
alexjfisher, line 13: Previously, this class needed to become the anchor resource.  By `require`ing it instead, we get the same effect.

PR #564, opened 2021-08-18
Remove explicit StandardOutput from webhook.service

kenyon:
systemd 247 complains that this setting is obsolete:

`systemd[1]: /etc/systemd/system/webhook.service:16: Standard output type syslog is obsolete, automatically updating to journal. Please update your unit file, and consider removing the setting altogether.`

The default for standard output is "journal" which means the same
thing as the old "syslog" setting.

### DIFF
--- a/templates/webhook.service.erb
+++ b/templates/webhook.service.erb
@@ -10,10 +10,7 @@ User=<%= @user %>
 TimeoutStartSec=90
 TimeoutStopSec=30
 RestartSec=10000
-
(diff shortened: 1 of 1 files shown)


smortex (approved):
I think it's fine.

Maybe just rename the issue something like "Remove explicit StandardOutput from webhook.service" so that users who have a custom `DefaultStandardOutput` in systemd are not surprised to see new messages there?  (I have not tested it but I assume this would change the behavior).

Issue #563, opened 2021-08-17
updates on existing unit file fail to restart service due to missing systemctl daemon-reload


tuxmea:
## How to reproduce (e.g Puppet code you use)

Deploy webhook with r10k v8.0.0 then switch to later version (we tested with v9.0.0).
Webhook service restart fails due to missing systemctl daemon-reload after unit file change

bastelfreak:
Hi,
is it possible that this happened on a Puppet 5 agent? We currently rely on puppet 6.1 or newer which has the systemctl daemon-reload builtin.

tuxmea:
Hi Tim,
used with latest v6 Puppet and with latest systemd module.

PR #566, opened 2021-08-19
Manage unit file via systemd::unit_file

bastelfreak:
fixes https://github.com/voxpupuli/puppet-r10k/issues/563


smortex (approved):
Looks good

PR #579, opened 2022-07-05
pool_size: default to amount of CPU cores

bastelfreak:

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -15,7 +15,7 @@ class r10k::params {
   $sources                = undef
   $puppet_master          = true
   $proxy                  = undef
-  $pool_size              = undef
+  $pool_size              = $facts['processors']['count']
   $gem_source             = undef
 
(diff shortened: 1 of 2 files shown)


manifests/params.pp
```diff
@@ -15,7 +15,7 @@
   $sources                = undef
   $puppet_master          = true
   $proxy                  = undef
-  $pool_size              = undef
+  $pool_size              = $facts['processors']['count']
```
rwaffen, line 18: often one does `cpu.count - 1` to not completely utilize the system. 
is this a thing we should consider here to?

bastelfreak: I don't think we need to consider it. My impression is that this is mostly IO bound or limited by the performance forge.puppet.com delivers.

rwaffen: okay. 👍  just came to my mind when i saw this change 🤔

PR #580, opened 2022-10-02
modulesync 5.3.0

bastelfreak:
modulesync 5.3.0

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -124,7 +124,7 @@ If you have Ruby 2.x or want a specific version of Puppet,
 you must set an environment variable such as:
 
 ```sh
-export PUPPET_VERSION="~> 5.5.6"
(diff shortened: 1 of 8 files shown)


bastelfreak:
@smortex this fails on Puppet 6 because some gems for the webhook require a newer ruby version. since puppet 6 is EoL soon anyway, I suggest to drop Puppet 6 in this module. Any objections?

smortex:
@bastelfreak I _think_ we can adjust something is `params.pp` to use on older sinatra when we are using an old version of Puppet:

https://github.com/voxpupuli/puppet-r10k/blob/b2951a6eac4bd022a5f1364c56971e01686e440c/manifests/params.pp#L156-L160

But I had no time to test it yesterday.  I'll try to give it a shot today, but if it's a dead end, I have no objection to stopping support for Puppet 6.

PR #585, opened 2023-01-13
Add the example gen-types script from the r10k FAQ.

jcpunk:
The `r10k` FAQ has a simple hint for how to automate generating types for environment isolation.  Seeding this into this module (but not folks systems by default) should make it easy for folks to quickly turn on isolation without much extra work.

The README is also updated with the relevant examples.


jcpunk:
Turns our r10k.yaml supports:
```yaml
deploy:
  generate_types: true
```

so this is somewhat pointless.

PR #594, opened 2023-05-04
Remove old webhook code in preparation for webhook-go being added

dhollinger:
Removed the old webhook script, puppet code, and tests in preparation for replacing it with new code for Webhook-go


.fixtures.yml
```diff
@@ -1,5 +1,12 @@
 fixtures:
   repositories:
+    stdlib: https://github.com/puppetlabs/puppetlabs-stdlib.git
```
bastelfreak, line 3: why do you add the modules?

dhollinger: That was a mistake from the merge conflicts. Should be fixed

bastelfreak (approved):

PR #605, opened 2023-06-09
Bugfix: Fix service name of webhook-go service

tmu-sprd:
The name of the systemd unit for the webhook service in the webhook-go package is named `webhook-go`. In the manifests and test, the legacy name `webhook` was still used. This PR fixes this.
Unit file at project voxpupuli/webhook-go: https://github.com/voxpupuli/webhook-go/blob/v2.1.0/build/webhook-go.service


bastelfreak (approved):

bastelfreak:
thanks for the fix! We should probably implement an acceptance test for the service.

PR #607, opened 2023-06-20
Bugfix: webhook: fix defaults, fix dependency, add acceptance test

tmu-sprd:
TL;DR
This PR fixes the defaults in params.pp for the webhook, adds a dependency to fix a first-run failure, adds a spec test for the defaults and an acceptance test.

Details:
When in a manifest classes are declared like this:
```puppet
class { 'r10k': }
-> class { 'r10k::webhook': }
```
a simple test with `it { is_expected.to compile.with_all_deps }` fails with:
```
  1) template on debian-10-x86_64 is expected to compile into a catalogue without dependency cycles
     Failure/Error: it { is_expected.to compile.with_all_deps }
     
       error during compilation: Evaluation Error: Error while evaluating a Resource Statement, Class[R10k::Webhook]:
         parameter 'chatops' entry 'service' expects an undef value or a match for Enum['rocketchat', 'slack'], got ''
         parameter 'chatops' entry 'channel' expects a value of type Undef or String[1], got String
         parameter 'chatops' entry 'user' expects a value of type Undef or String[1], got String
         parameter 'chatops' entry 'auth_token' expects a value of type Undef or String[1], got String
         parameter 'chatops' entry 'server_uri' expects a value of type Undef or String[1], got String
         parameter 'r10k' entry 'prefix' expects a value of type Undef or String[1], got String
         parameter 'config' entry 'chatops' entry 'service' expects an undef value or a match for Enum['rocketchat', 'slack'], got ''
         parameter 'config' entry 'chatops' entry 'channel' expects a value of type Undef or String[1], got String
         parameter 'config' entry 'chatops' entry 'user' expects a value of type Undef or String[1], got String
         parameter 'config' entry 'chatops' entry 'auth_token' expects a value of type Undef or String[1], got String
         parameter 'config' entry 'chatops' entry 'server_uri' expects a value of type Undef or String[1], got String
         parameter 'config' entry 'r10k' entry 'prefix' expects a value of type Undef or String[1], got String (file: /home/tmu/git/template/spec/fixtures/modules/template/manifests/init.pp, line: 6, column: 6) on node computer.example.com
     # ./spec/classes/init_spec.rb:21:in `block (4 levels) in <top (required)>'
```
Cause: The type is `Optional[Enum]` or `Optional[String[1]]`, but in params.pp the defaults for above values are empty strings `''`.

This PR changes those defaults to `undef`, which fixes this.

Additionally at the first Puppet run, Puppet fails with an error, because it tries to manages the file `/etc/voxpupuli/webhook.yml`, before the package `webhook-go` is being installed, which creates the directory `/etc/voxpupuli/`.
This PR adds the requirement to install the package to create the directory first before the file is being managed.

The unit test for the webhook is being changed to test, if the defaults do not fail.
An acceptance test is added to check, if the service is running with defaults.


tmu-sprd:
I have no idea, why the acceptance test for CentOS are failing. When run manually in a centos container, the error message is:
```
Error: Could not start Service[webhook-go]: Execution of '/sbin/service webhook-go start' returned 1: Redirecting to /bin/systemctl start webhook-go.service
Failed to get D-Bus connection: Operation not permitted
Error: /Stage[main]/R10k::Webhook::Service/Service[webhook-go]/ensure: change from 'stopped' to 'running' failed: Could not start Service[webhook-go]: Execution of '/sbin/service webhook-go start' returned 1: Redirecting to /bin/systemctl start webhook-go.service
Failed to get D-Bus connection: Operation not permitted
```
Google says, this may be caused by systemd in the container. I don't know, how to fix this and why it is working in other projects using the same setup (as far as I can see).

Archlinux is failing, because there is no support for it in beaker-puppet yet.

Any help is appreciated.

tmu-sprd:
The acceptance test at CentOS fails, because there is a redundant Alias entry in the systemd unit file shipped by the webhook-go package.
Created a PR to fix this: https://github.com/voxpupuli/webhook-go/pull/93
When this is released, the tests for CentOS will succeed.

sebastianrakel (approved):

PR #615, opened 2023-07-29
install.pp: Make class private; Update OpenBSD package name ruby22-r10k->ruby31-r10k

bastelfreak:


manifests/params.pp
```diff
@@ -1,7 +1,9 @@
 # Reasonable defaults for all classes
 class r10k::params {
-  $postrun                = undef
-  $package_name           = ''
+  $package_name = $facts['os']['name'] ? {
+    'OpenBSD' => 'ruby31-r10k',
```
ekohl, line 4: This is not bad, but it's a separate enhancement to mention.

bastelfreak: I updated the PR title.Do you think that's enough or should I pull it into a seperate PR?

jay7x (approved):

PR #631, opened 2023-09-05
webhook: Update default version 2.1.0->2.2.0

bastelfreak:
This fixes a bug in the systemd unit. That's a requirement to enable acceptance tests.


hashworks (approved):
LGTM

PR #633, opened 2023-09-05
webhook-go: Implement repository support

bastelfreak:


tuxmea (commented):
Should we set urls for repo installation via class parameter of should we have an option to completeley omit the installation so everybody can do the installation individually?

bastelfreak:
@tuxmea people can already configure if the repo should be configured or if another installation method should be used. I will add parameters for the URLs when I figure out the correct URLs. It's a bit try and error at the moment.

h-haaks:
I find the a bit strange that `install_method => 'none'` would actually install anything at all.

Isn't this better?
```
Enum['package', 'repo'] $install_method = 'package',
Boolean manage_repo = false,
```

PR #638, opened 2023-11-17
Update ordering on webhook and docu

### LINKED ISSUE #637: Webhook go ordering manages service prior installing package

- Puppet: any
- Ruby: any
- Distribution: any
- Module version: latest

## How to reproduce (e.g Puppet code you use)
```puppet
class { 'r10k':
  install_method => 'none',
}
include r10k::webhook
```
## What are you seeing

Service is managed prior package is installed.

## What behaviour did you expect instead

Package gets installed prior service is managed.

## Fix

order classes in `manifests/webhook.pp`

tuxmea:
webhook classes must be in order to ensure package and config prior service
update documentation on webhook-go curl command.

fixes #637

### DIFF
--- a/manifests/webhook.pp
+++ b/manifests/webhook.pp
@@ -53,4 +53,8 @@ class r10k::webhook (
   contain r10k::webhook::package
   contain r10k::webhook::config
   contain r10k::webhook::service
+
+  Class['r10k::webhook::package']
+  -> Class['r10k::webhook::config']
+  ~> Class['r10k::webhook::service']
 }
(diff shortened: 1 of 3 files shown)


manifests/webhook.pp
```diff
@@ -53,4 +53,8 @@
   contain r10k::webhook::package
   contain r10k::webhook::config
   contain r10k::webhook::service
+
+  Class['r10k::webhook::package']
+  -> Class['r10k::webhook::config']
+  ~> Class['r10k::webhook::service']
```
alexjfisher, line 59: I prefer the dependencies here, but this means we should remove this line? https://github.com/voxpupuli/puppet-r10k/blob/de72a78c1ceb32939a9928f9e5c3801db55ab904/manifests/webhook/config.pp#L10

tuxmea:
Can I please get a review?

rwaffen (approved):

PR #642, opened 2024-02-06
restore support for EoL Ruby / Puppet 7; enable acceptance tests

zilchms:
modulesync 7.3.0


zilchms:
we conclude, sometimes github is wierd. we changed nothing for the actual tests compared to #641, but now instead of 4 passing tests we have 6 failing ones, seems time to investigate....

PR #646, opened 2024-03-24
Add documentation stubs

zilchms:


.puppet-lint.rc
```diff
@@ -1,3 +1 @@
 --fail-on-warnings
---no-parameter_documentation-check
```
bastelfreak, line 2: Please also configure it in .sync.yml, otherwise modulesync will readd it.

zilchms: will do

PR #683, opened 2025-05-04
webhook: make blocked_branches options configureable

bastelfreak:
also contains https://github.com/voxpupuli/puppet-r10k/pull/682


bschonec:
@bastelfreak , could you add 'env_incremental' as well?  I submitted #701  but your changes are better.

PR #687, opened 2025-07-18
Remove dependency to mcollective module

h-haaks:

Using mcollective with r10k is optional.

Given the fact that the choria/mcollective module has not seen a new release in a year and by that does not support puppet/systemd 8.x, I think this module is better off having a 'soft' dependency to the module.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -5,7 +5,6 @@ fixtures:
     choria: https://github.com/choria-io/puppet-choria.git
     concat: https://github.com/puppetlabs/puppetlabs-concat.git
     cron_core: https://github.com/puppetlabs/puppetlabs-cron_core.git
-    git: https://github.com/puppetlabs/puppetlabs-git.git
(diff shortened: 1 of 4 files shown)


.fixtures.yml
```diff
@@ -5,7 +5,6 @@ fixtures:
     choria: https://github.com/choria-io/puppet-choria.git
     concat: https://github.com/puppetlabs/puppetlabs-concat.git
     cron_core: https://github.com/puppetlabs/puppetlabs-cron_core.git
-    git: https://github.com/puppetlabs/puppetlabs-git.git
```
bastelfreak, line 8: Can you please remove the other mcollective entries as well?

h-haaks: Tried that but then all the test failed ...
Had to put them back on by one to get all the tests green :)

TheMeier:
Hm, I thought optional dependencies are always listed since there is no way to express required vs optional in medata etc. This change is fine by me, but I wonder if this is another thing that should be discussed and documented in general.

